PR #2, opened 2016-12-11
Support UDP payloads and handlers

sempervictus:
With the addition of UDP sessions, this information needs to be
available to both L4 consumers. Technically speaking, socket.rb
can present an L3-only construct (raw IP sockets), which wont have
port information. However, this has not been an issue when using
the new transports in testing.

This is a breakout of the UDP handlers and payloads PR (#6035).

### DIFF
--- a/lib/rex/socket.rb
+++ b/lib/rex/socket.rb
@@ -746,6 +746,24 @@ module Socket
   end
 
   #
+  # Returns peer information (host + port) in host:port format.
+  #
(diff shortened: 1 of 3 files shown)


busterb:
This also turns off the forced IPv6 bind mode, which is good. I have a better patch for that though that I think mirrors the original intent.

bcook-r7:
Added fb195e7, cb39324

PR #5, opened 2017-06-25
Upstream socket.rb matchers

sempervictus:
Add regex for matching MAC addresses, as well as convenience
methods for checking if a string matches the regex, as well as a
generic :is_ip? method to verify if something is either IPv4 or v6.


sempervictus:
This is required for MSF PR #6611 upstreaming native DNS functionality.

Thanks @mubix for finding this, its invisible to my own tests since the forked gem has had this since prior to Rex separation efforts... can't actually trace the original date.

PR #8, opened 2017-12-28
Extract and mixin cert ops from server module

sempervictus:
Generic SSL routines can be in their own module, for import by
consumers without having to drag the entire server infrastructure
in with it.

This pulls the certificate methods into Rex::Socket::Ssl for use
by consumers, and includes the module in Rex::Socket::SslTcpServer
as the initial consumer.

### DIFF
--- a/lib/rex/socket/ssl.rb
+++ b/lib/rex/socket/ssl.rb
@@ -19,6 +19,65 @@ module Rex::Socket::Ssl
   rescue ::Exception
   end
 
+  module CertProvider
+
+    def self.ssl_generate_subject
+      st  = Rex::Text.rand_state
+      loc = Rex::Text.rand_name.capitalize
+      org = Rex::Text.rand_name.capitalize
+      cn  = Rex::Text.rand_hostname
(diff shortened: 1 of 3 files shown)


lib/rex/socket/ssl.rb
```diff
@@ -0,0 +1,148 @@
...
+###
+#
+# This class provides methods for interacting with an SSL wrapped TCP server.  It
+# implements the StreamServer IO interface.
+#
+###
+module Rex::Socket::Ssl
+
+  @@loaded_openssl = false
+
```
busterb: do you think we really need to continue pretending openssl is optional? I don't think Ruby even compiles without it anymore. I'd suggest removing this entire @@loaded_openssl path

sempervictus: Done sir done.

sempervictus:
Well, apprently New England and the midwest arent the only places frozen over. Hell's boiler must be broken because these all claim to pass CI. :)

busterb:
Yeah, that's a good sign Thanks!

PR #12, opened 2018-03-30
Add IPv6 Support To addr_atoc And addr_ctoa

zeroSteiner:
This PR addess support for IPv6 addresses in the `addr_atoc` and `addr_ctoa` functions. This is necessary for a PR I'll be submitting to the Metasploit Framework side of things to use these functions to restore IPv6 support for the `route` command.

Example Usage:
```
>> Rex::Socket.addr_ctoa(32, v6=true)
=> "ffff:ffff::"
>> Rex::Socket.addr_ctoa(32, v6=false)
=> "255.255.255.255"
>> Rex::Socket.addr_atoc('ffff:ffff::')
=> 32
>> Rex::Socket.addr_atoc('255.255.255.255')
=> 32
>> Rex::Socket.addr_ctoa(38, v6=true)
=> "ffff:ffff:fc00::"
>> Rex::Socket.addr_atoc('ffff:ffff:fc00::')
=> 38
>> 
```

The PR which will require this will actually leverage the functionality and will be better for trying to test this out.

### DIFF
--- a/lib/rex/socket.rb
+++ b/lib/rex/socket.rb
@@ -289,10 +289,11 @@ module Socket
   # nil if it's not convertable.
   #
   def self.addr_atoc(mask)
+    bits = is_ipv6?(mask) ? 128 : 32
     mask_i = resolv_nbo_i(mask)
     cidr = nil
-    0.upto(32) do |i|
-      if ((1 << i)-1) << (32-i) == mask_i
+    0.upto(bits) do |i|
+      if ((1 << i)-1) << (bits - i) == mask_i
         cidr = i
         break
       end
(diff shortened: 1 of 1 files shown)


acammack-r7:
Nifty! One small thing, the proper syntax for named variables in Ruby is: `foo(bar: true)`, in your code snippet above, you are assigning a value to the variable `v6` during the function call, which returns its value so it gets passed to the optional positional argument. To make a keyword argument, you need `def foo(bar: false)`. On a personal note, I would name the option `ip6`.

zeroSteiner:
Well I got a little carried away and broke the unit tests. Bare with me while I read up on proper Ruby 2 keyword argument definition / usage.

acammack-r7:
For methods where you just pass it through, you can `**splat` them just like Python.

zeroSteiner:
Ok, so I reverted the broken overzealous commit and `addr_ctoa` now uses a new-style keyword argument.

PR #13, opened 2018-07-25
SSH socket registration removed

sgonzalez-r7:
When using SSHFactory to create a socket to be used by net-ssh, that socket was registered, presumably for later cleanup.  The registration, however, was leading to premature socket closure and causing bad file descriptor errors downstream in an exploit that uses net-ssh.  This PR removes such registration.

### DIFF
--- a/lib/rex/socket/ssh_factory.rb
+++ b/lib/rex/socket/ssh_factory.rb
@@ -38,7 +38,6 @@ module Rex
             'MsfExploit'   => msfmodule
           }
         )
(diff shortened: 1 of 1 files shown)


busterb:
at this point, we also don't even need msfmodule anymore, or a lot of other metasploit-leakage into the rex library

busterb:
I'll leave those in place for now though. It's surprising this in rex, considering how Metasploit-specific it is.

PR #16, opened 2019-02-27
Bring RangeWalker to IPv6 land

acammack-r7:
Adds CIDR (new) and range (broke in 2011) support to IPv6 ranges.

Verification
========

- [x] `rake spec` should pass
- [x] `ruby bin/console`
- [x] `Rex::Socket::RangeWalker.new` should work with:
  - [x] Ranges like `'fe80::3330-fe80::00ab:f000'`
  - [x] And CIDRs like `'fe80::/16'`
  - [x] And stuff with scopes like `'fe80::%eth0/100'`

### DIFF
--- a/lib/rex/socket/range_walker.rb
+++ b/lib/rex/socket/range_walker.rb
@@ -97,8 +97,9 @@ class RangeWalker
 
         addr1, scope_id = addrs[0].split('%')
         opts[:scope_id] = scope_id if scope_id
+        opts[:ipv6] = true
 
-        addr2, scope_id = addrs[0].split('%')
+        addr2, scope_id = addrs[1].split('%')
         ( opts[:scope_id] ||= scope_id ) if scope_id
 
         # Both have to be IPv6 for this to work
--- a/lib/rex/socket/range_walker.rb
+++ b/lib/rex/socket/range_walker.rb
@@ -81,7 +81,7 @@ class RangeWalker
 
       # Handle IPv6 CIDR first
       if arg.include?(':') && arg.include?('/')
-        return false if !valid_cidr_chars(arg)
+        return false if !valid_cidr_chars?(arg)
 
         ip_part,mask_part = arg.split("/")
         return false unless (0..128).include? mask_part.to_i
@@ -128,7 +128,7 @@ class RangeWalker
       # Handle IPv4 CIDR
       elsif arg.include?("/")
         # Then it's CIDR notation and needs special case
-        return false if !valid_cidr_chars(arg)
+        return false if !valid_cidr_chars?(arg)
         ip_part,mask_part = arg.split("/")
         return false unless (0..32).include? mask_part.to_i
         if ip_part =~ /^\d{1,3}(\.\d{1,3}){1,3}$/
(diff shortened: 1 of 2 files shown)


lib/rex/socket/range_walker.rb
```diff
@@ -79,8 +79,21 @@ def parse(parseme)
     parseme.split(', ').map{ |a| a.split(' ') }.flatten.each do |arg|
       opts = {}
 
-      # Handle IPv6 first (support ranges, but not CIDR)
-      if arg.include?(":")
+      # Handle IPv6 CIDR first
+      if arg.include?(':') && arg.include?('/')
+        return false if !valid_cidr_chars(arg)
```
wvu-r7: `if !` vs. `unless` below.

lib/rex/socket/range_walker.rb
```diff
@@ -79,8 +79,21 @@ def parse(parseme)
...
+
+        ip_part,mask_part = arg.split("/")
```
wvu-r7: Space after comma.

lib/rex/socket/range_walker.rb
wvu-r7: I think the way you have it might read better, at least for me.

lib/rex/socket/range_walker.rb
```diff
@@ -417,6 +428,17 @@ def expand_nmap(arg)
     return ranges
   end
 
+  protected
+
+  def valid_cidr_chars(arg)
```
wvu-r7: This looks like it can be `valid_cidr_chars?`, since you return either `true` or `false` and nothing else.

lib/rex/socket/range_walker.rb
```diff
@@ -417,6 +428,17 @@ def expand_nmap(arg)
...
+    return false if arg.include? ',-' # Improper CIDR notation (can't mix with 1,3 or 1-3 style IP ranges)
+    return false if arg.scan("/").size > 1 # ..but there are too many slashes
+    ip_part,mask_part = arg.split("/")
+    return false if ip_part.nil? or ip_part.empty? or mask_part.nil? or mask_part.empty?
```
wvu-r7: `or`?

acammack-r7: Line was copy pasta'd, I can bring it to the future as well.

lib/rex/socket/range_walker.rb
acammack-r7: Good point, I'll update.

lib/rex/socket/range_walker.rb
acammack-r7: With these sorts of interrogators I feel `if !` reads better. If you come up with a name that reads well with `unless` I'l be happy to change it.

wvu-r7: Nope, we're in agreement.

pbarry-r7:
It does pass the `rake spec` tests and the above 3 `Rex::Socket::RangeWalker.new` steps do work for me.

Question/observation: is the `fe80::3330-fe80::00ab:f000` example a valid IPv6 range?  I believe the presence of two double colons isn't allowed (in https://en.wikipedia.org/wiki/IPv6_address#Representation, see the line which starts "One or more consecutive groups...").  The RangeWalker accepts that `fe80::3330-fe80::00ab:f000` without reporting an error, but balks if I give it a range that I believe is valid (please do correct me if I'm incorrect here):

```
2.5.3 :019 > Rex::Socket::RangeWalker.new("fe80::3330-fe80::00ab:f000")
 => #<Rex::Socket::RangeWalker:0x0000000001936330 @ranges=[#<Rex::Socket::Range:0x0000000001935688 @start=338288524927261089654018896841347707696, @stop=338288524927261089654018896841358962688, @options={:ipv6=>true}>], @curr_range_index=0, @curr_addr=338288524927261089654018896841347707696, @length=11254993> 
2.5.3 :020 > 
2.5.3 :021 > Rex::Socket::RangeWalker.new("fe80::3330-fe80:00ab:f000")
 => #<Rex::Socket::RangeWalker:0x000000000193fde0 @ranges=false> 
2.5.3 :022 > Rex::Socket::RangeWalker.new("fe80:3330-fe80::00ab:f000")
 => #<Rex::Socket::RangeWalker:0x0000000001947130 @ranges=false> 
2.5.3 :023 > Rex::Socket::RangeWalker.new("fe80:0000:0000:0000:0000:3330-fe80:00ab:f000")
 => #<Rex::Socket::RangeWalker:0x0000000001432ca8 @ranges=false> 
```

acammack-r7:
The parts separated by a `-` are two distinct IP addresses. This doesn't add nmap-style expansion, only `begin-end` style.

PR #17, opened 2019-06-24
Set certificate hostname from params

sempervictus:
This should help a bit with making the certificate common name match the VHOST or explicit SSLCommonName param passed down from Msf.
Not tested yet, needed for CONNECT->TLS MITM in the HTTP proxy (i think), should be useful in other spots too.


busterb:
Looks good to me, thanks @sempervictus

PR #19, opened 2019-08-30
Unify SSL cert provider with the one in framework

acammack-r7:
These interfaces were expanded in incompatible directs, so unify around
keyword arguments.

This modifies the interface changes in #17/#18 to be more flexible and not clash with the SSL certificate provider in framework.

Verification
========

- [x] Check out https://github.com/rapid7/metasploit-framework/pull/12251 and point your `Gemfile` to use this branch (or your local checkout of it) for `rex-socket`
- [x] Start framework
- [x] Verify that you can use handlers that need SSL certs

Landing
======

Thanks to the interdependent nature of the fix the order should be:
- [ ] Land this
- [ ] Wait for the new version to publish
- [ ] Bump the `rex-text` version in `Gemfile.lock` for https://github.com/rapid7/metasploit-framework/pull/12251 or let me update it
- [ ] Land https://github.com/rapid7/metasploit-framework/pull/12251

### DIFF
--- a/lib/rex/socket/ssl.rb
+++ b/lib/rex/socket/ssl.rb
@@ -13,10 +13,12 @@ module Rex::Socket::Ssl
 
   module CertProvider
 
-    def self.ssl_generate_subject(cn = Rex::Text.rand_hostname, org = Rex::Text.rand_name)
-      st  = Rex::Text.rand_state
-      loc = Rex::Text.rand_name.capitalize
-      "/C=US/ST=#{st}/L=#{loc}/O=#{org.capitalize}/CN=#{cn}"
+    def self.ssl_generate_subject(cn: nil, org: nil, loc: nil, st: nil)
+      st  ||= Rex::Text.rand_state
+      loc ||= Rex::Text.rand_name.capitalize
+      org ||= Rex::Text.rand_name.capitalize
+      cn  ||= Rex::Text.rand_hostname
+      "/C=US/ST=#{st}/L=#{loc}/O=#{org}/CN=#{cn}"
     end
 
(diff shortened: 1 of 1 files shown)


lib/rex/socket/ssl.rb
```diff
@@ -13,10 +13,12 @@ module Rex::Socket::Ssl
...
-    def self.ssl_generate_subject(cn = Rex::Text.rand_hostname, org = Rex::Text.rand_name)
-      st  = Rex::Text.rand_state
-      loc = Rex::Text.rand_name.capitalize
-      "/C=US/ST=#{st}/L=#{loc}/O=#{org.capitalize}/CN=#{cn}"
+    def self.ssl_generate_subject(cn: nil, org: nil, loc: nil, st: nil)
+      st  ||= Rex::Text.rand_state
+      loc ||= Rex::Text.rand_name.capitalize
+      org ||= Rex::Text.rand_name.capitalize
+      cn  ||= Rex::Text.rand_hostname
+      "/C=US/ST=#{st}/L=#{loc}/O=#{org}/CN=#{cn}"
```
busterb: At first I thought "Hmm, why not expand this to country", but figured that then having to deal with state/country mismatch wouldn't be fun.

acammack-r7: Yeah, I'd like to have less duplication between this and https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/cert_provider.rb overall, but I figured I'd leave it at getting them to equivalent-ish interfaces for now.

sempervictus:
Thank you!

PR #22, opened 2020-02-17
Add to_ and merge_ hash methods to the Parameters class

zeroSteiner:
This PR adds two new methods (and some basic testing) to the `Rex::Socket::Parameters` class. The intention is to allow the class to be updated while leveraging the existing logic within the `#new` method. This will be used in a PR I'll be submitting soon to Metasploit to allow parameters to be updated with reported `LocalHost` and `LocalPort` settings for sockets that are determined by the OS (as is the case when `LocalPort` is `0` for example).

## Testing Steps
- [ ] Make sure the new rspec tests pass and look reasonable
- [ ] Check the new methods, something similar to the example below

## Example Usage
```
msf5 > irb
[*] Starting IRB shell...

>> params = Rex::Socket::Parameters.new({})
=> #<Rex::Socket::Parameters:0x00007f8e284f1240 @peerhost=nil, @localhost="0.0.0.0", @peerport=0, @localport=0, @bare=false, @ssl=false, @proto="tcp", @server=false, @comm=Rex::Socket::Comm::Local, @context={}, @retries=0, @timeout=5, @v6=false>
>> params.to_hash
=> {"LocalHost"=>"0.0.0.0", "PeerPort"=>0, "LocalPort"=>0, "Bare"=>false, "SSL"=>false, "Proto"=>"tcp", "Server"=>false, "Comm"=>Rex::Socket::Comm::Local, "Context"=>{}, "Retries"=>0, "Timeout"=>5, "IPv6"=>false}
>> params.merge_hash({"LocalHost" => "127.0.0.1", "LocalPort" => 51738})
=> #<Rex::Socket::Parameters:0x00007f8e290d1f88 @peerhost=nil, @localhost="127.0.0.1", @peerport=0, @localport=51738, @bare=false, @ssl=false, @proto="tcp", @server=false, @comm=Rex::Socket::Comm::Local, @context={}, @retries=0, @timeout=5, @v6=false>
>> params.merge_hash({"LocalHost" => "127.0.0.1", "LocalPort" => 51738}).to_hash
=> {"LocalHost"=>"127.0.0.1", "PeerPort"=>0, "LocalPort"=>51738, "Bare"=>false, "SSL"=>false, "Proto"=>"tcp", "Server"=>false, "Comm"=>Rex::Socket::Comm::Local, "Context"=>{}, "Retries"=>0, "Timeout"=>5, "IPv6"=>false}
>> params.merge_hash({"LocalHost" => "127.0.0.1", "LocalPort" => 51738}).to_hash
```

### DIFF
--- a/lib/rex/socket/parameters.rb
+++ b/lib/rex/socket/parameters.rb
@@ -329,15 +329,16 @@ class Rex::Socket::Parameters
   def comm
     return @comm unless @comm.nil?
 
+    best_comm = nil
     # If no comm was explicitly specified, try to use the comm that is best fit
     # to handle the provided host based on the current routing table.
     if server and localhost
-      return Rex::Socket::SwitchBoard.best_comm(localhost)
+      best_comm = Rex::Socket::SwitchBoard.best_comm(localhost)
     elsif peerhost
-      return Rex::Socket::SwitchBoard.best_comm(peerhost)
+      best_comm =  Rex::Socket::SwitchBoard.best_comm(peerhost)
     end
 
-    Rex::Socket::Comm::Local
+    best_comm || Rex::Socket::Comm::Local
   end
 
   # The context hash that was passed in to the structure.  (default: {})
--- a/lib/rex/socket/parameters.rb
+++ b/lib/rex/socket/parameters.rb
@@ -154,6 +154,8 @@ class Rex::Socket::Parameters
       rescue ::Exception => e
         elog("Failed to read cert: #{e.class}: #{e}", LogSource)
       end
+    elsif hash['SSLCertRaw']
+      self.ssl_cert = hash['SSLCertRaw']
     end
 
     if (hash['SSLClientCert'] and ::File.file?(hash['SSLClientCert']))
@@ -162,6 +164,8 @@ class Rex::Socket::Parameters
       rescue ::Exception => e
         elog("Failed to read client cert: #{e.class}: #{e}", LogSource)
       end
+    elsif hash['SSLClientCertRaw']
+      self.ssl_client_cert = hash['SSLClientCertRaw']
     end
 
     if (hash['SSLClientKey'] and ::File.file?(hash['SSLClientKey']))
@@ -170,6 +174,8 @@ class Rex::Socket::Parameters
       rescue ::Exception => e
         elog("Failed to read client key: #{e.class}: #{e}", LogSource)
       end
+    elsif hash['SSLClientKeyRaw']
+      self.ssl_client_key = hash['SSLClientKeyRaw']
     end
 
     if hash['Proxies']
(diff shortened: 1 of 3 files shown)


zeroSteiner:
It looks like there's an issue with the Bundler version that's breaking the Travis build.

lib/rex/socket/parameters.rb
```diff
@@ -231,6 +237,41 @@ def initialize(hash)
...
+    settings['Comm'] = self.comm
+    settings['Context'] = self.context
+    settings['Retries'] = self.retries
+    settings['Timeout'] = self.timeout
+    settings['IPv6'] = self.v6
+
+    settings
+  end
+
+  def merge_hash(other_hash)
```
acammack-r7: Serialization is nice, but I think this would be better as a plain `merge` method that may or may not use hashes under the covers. Is there an anticipated need for `to_hash` aside from this method? If not, I would consider moving most of the code in `initialize` to something like `import_settings_from_hash` and have this method go like:
```ruby
def merge(other)
  case other
  when Hash
    import_settings_from_hash(other)
  when Parameters
    # import straight across if needed, possibly using metaprogramming
    other.instance_variables.each do |var|
      val = other.instance_variable_get(var)
      instance_variable_set(var, val) unless val.nil?
    end
  end
end
```
To accommodate this, the new `import_settings_from_hash` method would need to start checking for `self.xxxxx` when setting defaults, but I think that still might be better than maintaining the list of params and their hash keys in two different spots. Alternately, you could try untangling the bulk import from the default setting code, possibly moving some of the defaults into manually written accessor methods.

acammack-r7: Mostly, I feel having two separate hash key -> variable methods this large is a recipe for bugs. I'm open to other alternatives.

zeroSteiner: My concern with that approach was handling the few settings that have defaults that are not nil like localhost, localport and peerport, ssl. We could handle those seperately but I'm not sure how we'd track the difference between explicitly setting localport to 0 and a localport that was left at it's default value of 0 and thus shouldn't be updated.

acammack-r7: That's where the custom accessors could come into play:
```ruby
attr_writer :localport
def localport
  @localport || 0
end
```
Doing the instance variable merge would then not overwrite things with default values from unset variables. This approach would also help with the serialize-deserialize approach currently here, which has the same issue.

spec/rex/socket/parameters_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  describe '#to_hash' do
+    it "should return a populated hash" do
+      params_hash = params.to_hash
+      expect(params_hash).to be_a(Hash)
+      expect(params_hash.length).to be > 0
+      expect(params_hash).to include("LocalHost", "LocalPort", "IPv6", "SSL", "Server")
+    end
+  end
```
acammack-r7: Just a reminder to update these tests.

busterb:
all green now with an updated Ruby

PR #24, opened 2020-09-05
x509_certificate: add use of EC keys

bigbael:
Add new capability to use EC keys,  Resolves issue with using EC keys in metasploit.


bigbael:
will request again shortly.

PR #25, opened 2020-09-09
x509_certificate: Allow use of EC keys

bigbael:
Add new capability to allow the use of EC keys.  Resolves issue with using EC keys in metasploit.

### DIFF
--- a/lib/rex/socket/x509_certificate.rb
+++ b/lib/rex/socket/x509_certificate.rb
@@ -28,7 +28,9 @@ class X509Certificate
 
     certs = []
     ssl_cert.scan(/-----BEGIN\s*[^\-]+-----+\r?\n[^\-]*-----END\s*[^\-]+-----\r?\n?/nm).each do |pem|
-      if pem =~ /PRIVATE KEY/
+      if pem =~ /EC PRIVATE KEY/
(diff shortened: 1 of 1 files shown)


smcintyre-r7:
> Resolves issue with using EC keys in metasploit.

Can you please provide additional detail (and ideally steps) that we can use to reproduce the issue that this resolves?

bigbael:
To reproduce the issue, use the following steps, from msfconsole:
use exploit/multi/handler
set payload windows/meterpreter/reverse_https (or something using https)
set necessary lport and lhost (0.0.0.0 and 443 will work)
set handlersslcert <path to EC key cert>

Attempt to start this handler, you will get a bind error due to inability to read the EC private key.

With patch, EC key is handled and listener starts with appropriate config.

smcintyre-r7:
Alright I tested this out and was able to reproduce the original issue and then verify that the patch fixes it. I used the steps from [here](https://www.guyrutenberg.com/2013/12/28/creating-self-signed-ecdsa-ssl-certificate-using-openssl/) to generate a certificate for testing and then used the `exploit/multi/handler` module.

I'll get this merged in momentarily.

smcintyre-r7 (approved):

PR #27, opened 2021-02-08
Add github actions for testing and some changes to support Ruby 3

cgranleese-r7:
This PR adds some changes that were required to transition over to `Ruby 3`, as well as adding GitHub actions for testing.

Adding GitHub actions for testing:
image

This PR fixes the following Ruby 3 deprecation warnings:
image

## Note
Some tests were removed, the previous `::Socket.gethostbyname()` returned an array of information, which included the byte address associated with `host/hostname`. The new ::Addrinfo.getaddrinfo() now returns an array of addresses only.

Any other code that was dependant on `::Socket.gethostbyname()` returning the byte addresses has been changed to handle that with a separate method call. 

Ruby 3 upgrade ticket https://github.com/rapid7/metasploit-framework/issues/14666

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,4 @@ source 'https://rubygems.org'
 
 # Specify your gem's dependencies in rex-socket.gemspec
 gemspec
+gem 'pry', '~> 0.13.1'
--- a/Gemfile
+++ b/Gemfile
@@ -2,4 +2,7 @@ source 'https://rubygems.org'
 
 # Specify your gem's dependencies in rex-socket.gemspec
 gemspec
-gem 'pry', '~> 0.13.1'
+
+group :development do
+  gem 'pry-byebug'
+end
--- a/lib/rex/socket.rb
+++ b/lib/rex/socket.rb
@@ -192,15 +192,8 @@ module Socket
       address_info.ip_address
     end
 
-    if res[0] =~ MATCH_IPV4 || res[0] =~ MATCH_IPV6
-      unless accept_ipv6
-        res.reject!{ |ascii| ascii =~ MATCH_IPV6 }
-      end
-    else
-      unless accept_ipv6
-        res.reject!{ |nbo| nbo.length != 4 }
-      end
-      res.map!{ |nbo| self.addr_ntoa(nbo) }
+    unless accept_ipv6
+      res.reject! { |ascii| ascii =~ MATCH_IPV6 }
     end
 
     res
@@ -212,7 +205,9 @@ module Socket
   # not occur.  This is done in order to prevent delays, such as would occur
   # on Windows.
   #
+  # @deprecated Please use {#getaddress}, {#resolv_nbo}, or similar instead.
   def self.gethostbyname(host)
+    warn "NOTE: #{self}.#{__method__} is deprecated, use getaddress, resolve_nbo, or similar instead. It will be removed in the next Major version"
     if is_ipv4?(host)
       return [ host, [], 2, host.split('.').map{ |c| c.to_i }.pack("C4") ]
     end
@@ -223,12 +218,7 @@ module Socket
       host, _ = host.split('%', 2)
     end
 
-    res = ::Addrinfo.getaddrinfo(host, 0, ::Socket::AF_UNSPEC, ::Socket::SOCK_STREAM)
-    res.map! do |address_info|
-      address_info.ip_address
-    end
-
-    res
+    ::Socket.gethostbyname(host)
   end
 
   #
@@ -259,8 +249,8 @@ module Socket
   #
   # Resolves a host to raw network-byte order.
   #
-  def self.resolv_nbo(host)
-    ip_address = Rex::Socket.getaddress(host)
+  def self.resolv_nbo(host, accepts_ipv6 = true)
+    ip_address = Rex::Socket.getaddress(host, accepts_ipv6)
     IPAddr.new(ip_address).hton
   end
 
--- a/lib/rex/socket.rb
+++ b/lib/rex/socket.rb
@@ -186,14 +186,12 @@ module Socket
       return [hostname]
     end
 
-    res = ::Socket.gethostbyname(hostname)
-    return [] if not res
+    res = ::Addrinfo.getaddrinfo(hostname, 0, ::Socket::AF_UNSPEC, ::Socket::SOCK_STREAM)
 
-    # Shift the first three elements out, leaving just the list of
-    # addresses
-    res.shift # name
-    res.shift # alias hostnames
-    res.shift # address_family
+    res.map! do |address_info|
+      address_info.ip_address
+    end
+    return [] if not res
 
     # Rubinius has a bug where gethostbyname returns dotted quads instead of
     # NBO, but that's what we want anyway, so just short-circuit here.
@@ -228,7 +226,12 @@ module Socket
       host, _ = host.split('%', 2)
     end
 
-    ::Socket.gethostbyname(host)
+    res = ::Addrinfo.getaddrinfo(host, 0, ::Socket::AF_UNSPEC, ::Socket::SOCK_STREAM)
+    res.map! do |address_info|
+      address_info.ip_address
+    end
+
+    res
   end
 
   #
@@ -260,14 +263,17 @@ module Socket
   # Resolves a host to raw network-byte order.
   #
   def self.resolv_nbo(host)
-    self.gethostbyname( Rex::Socket.getaddress(host, true) )[3]
+    ip_address = Rex::Socket.getaddress(host)
+    IPAddr.new(ip_address).hton
   end
 
   #
   # Resolves a host to raw network-byte order.
   #
   def self.resolv_nbo_list(host)
-    Rex::Socket.getaddresses(host).map{|addr| self.gethostbyname(addr)[3] }
+    Rex::Socket.getaddresses(host).map do |addresses|
+      IPAddr.new(addresses).hton
+    end
   end
 
   #
--- a/lib/rex/socket.rb
+++ b/lib/rex/socket.rb
@@ -174,9 +174,10 @@ module Socket
   end
 
   #
-  # ::Addrinfo.getaddrinfo checks to see if the supplied address is already
-  # an ASCII IP address.  This is necessary to prevent blocking while waiting
-  # on a DNS reverse lookup when we already have what we need.
+  # Wrapper for +::Addrinfo.getaddrinfo+ that takes special care to see if the
+  # supplied address is already an ASCII IP address.  This is necessary to
+  # prevent blocking while waiting on a DNS reverse lookup when we already
+  # have what we need.
   #
   # @param hostname [String] A hostname or ASCII IP address
   # @return [Array<String>]
@@ -190,7 +191,17 @@ module Socket
     res.map! do |address_info|
       address_info.ip_address
     end
-    return [] if not res
+
+    if res[0] =~ MATCH_IPV4 || res[0] =~ MATCH_IPV6
+      unless accept_ipv6
+        res.reject!{ |ascii| ascii =~ MATCH_IPV6 }
+      end
+    else
+      unless accept_ipv6
+        res.reject!{ |nbo| nbo.length != 4 }
+      end
+      res.map!{ |nbo| self.addr_ntoa(nbo) }
+    end
 
     res
   end
--- a/lib/rex/socket.rb
+++ b/lib/rex/socket.rb
@@ -174,10 +174,9 @@ module Socket
   end
 
   #
-  # Wrapper for +::Socket.gethostbyname+ that takes special care to see if the
-  # supplied address is already an ASCII IP address.  This is necessary to
-  # prevent blocking while waiting on a DNS reverse lookup when we already
-  # have what we need.
+  # ::Addrinfo.getaddrinfo checks to see if the supplied address is already
+  # an ASCII IP address.  This is necessary to prevent blocking while waiting
+  # on a DNS reverse lookup when we already have what we need.
   #
   # @param hostname [String] A hostname or ASCII IP address
   # @return [Array<String>]
@@ -192,19 +191,6 @@ module Socket
       address_info.ip_address
     end
     return [] if not res
-
-    # Rubinius has a bug where gethostbyname returns dotted quads instead of
-    # NBO, but that's what we want anyway, so just short-circuit here.
-    if res[0] =~ MATCH_IPV4 || res[0] =~ MATCH_IPV6
-      unless accept_ipv6
-        res.reject!{ |ascii| ascii =~ MATCH_IPV6 }
-      end
-    else
-      unless accept_ipv6
-        res.reject!{ |nbo| nbo.length != 4 }
-      end
-      res.map!{ |nbo| self.addr_ntoa(nbo) }
-    end
 
     res
   end
--- a/spec/rex/socket_spec.rb
+++ b/spec/rex/socket_spec.rb
@@ -110,7 +110,8 @@ RSpec.describe Rex::Socket do
 
   describe '.getaddresses' do
 
-    subject { described_class.getaddresses('whatever') }
+    let(:accepts_ipv6) { true }
+    subject { described_class.getaddresses('whatever', accepts_ipv6) }
 
     before(:example) do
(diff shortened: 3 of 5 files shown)


spec/rex/socket_spec.rb
```diff
@@ -150,19 +130,6 @@
...
-      let(:response_addresses) { ["1.1.1.1", "2.2.2.2"] }
-
-      it { is_expected.to be_an(Array) }
-      it { expect(subject.size).to eq(2) }
-      it "should return the ASCII addresses" do
-        expect(subject).to include("1.1.1.1")
-        expect(subject).to include("2.2.2.2")
-      end
-
-    end
```
adfoster-r7, line 165: Is this rubinus bug fixed now?

spec/rex/socket_spec.rb
```diff
@@ -84,27 +84,7 @@
...
-
-    context 'when ::Socket.gethostbyname returns IPv6 responses' do
-      let(:response_afamily) { Socket::AF_INET6 }
-      let(:response_addresses) { ["\xfe\x80"+("\x00"*13)+"\x01", "\xfe\x80"+("\x00"*13)+"\x02"] }
-
-      it { is_expected.to be_a(String) }
-      it "should return the first ASCII address" do
-        expect(subject).to eq "fe80::1"
-      end
```
adfoster-r7, line 107: Will this break framework if these don't pass anymore?

cgranleese-r7: _My understanding_, is that `::Socket.gethostbyname` returned an array with several pieces of information, which inlcluded the byte address. Now we're using `::Addrinfo.getaddrinfo`, which only returns the ip address, so these tests are now redundant.

The tests that expect `"1.1.1.1" & "fe80::1"` to be returned are already handled in other places. (Line 112 & 124)

_But_, this comment made me realise I didn't change the context description on those two tests, so I'll get that sorted.

lib/rex/socket.rb
```diff
@@ -228,7 +226,12 @@ def self.gethostbyname(host)
       host, _ = host.split('%', 2)
     end
 
-    ::Socket.gethostbyname(host)
+    res = ::Addrinfo.getaddrinfo(host, 0, ::Socket::AF_UNSPEC, ::Socket::SOCK_STREAM)
+    res.map! do |address_info|
+      address_info.ip_address
+    end
+
+    res
```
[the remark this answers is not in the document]
adfoster-r7: Taking a step back, if this file is just a wrapper around Ruby's `::Socket`, and Ruby themselves have marked `Socket::gethostbyname` as deprecated, maybe we need to do the same - and add an additional method for the newer API, and update metasploit-framework to align with this new API

This is potentially more effort, but potentially worth considering the scope of such a change

spec/rex/socket_spec.rb
adfoster-r7: I think these tests need updated similar to how `.getaddresses` was - rather than updated 👀

spec/rex/socket_spec.rb
```diff
@@ -84,27 +84,7 @@
...
+      allow(Addrinfo).to receive(:getaddrinfo).and_return(response_addresses.map {|address| Addrinfo.ip(address)})
     end
 
     context "with rubinius' bug returning ASCII addresses" do
```
adfoster-r7, line 90: Same as below 👍

lib/rex/socket.rb
```diff
@@ -186,14 +186,12 @@ def self.getaddresses(hostname, accept_ipv6 = true)
...
 
-    # Shift the first three elements out, leaving just the list of
-    # addresses
-    res.shift # name
-    res.shift # alias hostnames
-    res.shift # address_family
+    res.map! do |address_info|
+      address_info.ip_address
+    end
+    return [] if not res
```
adfoster-r7, line 194: This seems like an implementation hangover that could probably be deleted once the below comment is resolved

lib/rex/socket.rb
```diff
@@ -260,14 +263,17 @@ def self.from_sockaddr(saddr)
...
   end
 
   #
   # Resolves a host to raw network-byte order.
   #
   def self.resolv_nbo_list(host)
-    Rex::Socket.getaddresses(host).map{|addr| self.gethostbyname(addr)[3] }
+    Rex::Socket.getaddresses(host).map do |addresses|
+      IPAddr.new(addresses).hton
+    end
```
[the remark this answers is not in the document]
adfoster-r7: Looking closer though, it seems like this function isn't an exact wrapper around `resolv_nbo` ? 👀

spec/rex/socket_spec.rb
cgranleese-r7: Sure, I'll get those added back in and updated 👍

spec/rex/socket_spec.rb
cgranleese-r7: Going by these comments below, the Rubinius bug has been removed due to it been tied to the deprecated method (`::Socket.gethostbyname`). 
```Ruby
# Rubinius has a bug where gethostbyname returns dotted quads instead of
# NBO, but that's what we want anyway, so just short-circuit here.
```

lib/rex/socket.rb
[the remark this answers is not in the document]
cgranleese-r7: So here, `self.gethostbyname(host)` is only called once to return the byte address here:
```Ruby
def self.resolv_nbo(host)
  self.gethostbyname( Rex::Socket.getaddress(host, true) )[3]
end
```
Before the only information being used was the 4th element of the array:
```
["localhost", [], 30, "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01", "\x7F\x00\x00\x01"]
```
So as our new functionality returns IP addresses, we added this new code to convert the returned value to a byte address.
```Ruby
def self.resolv_nbo(host)
  ip_address = Rex::Socket.getaddress(host)
  IPAddr.new(ip_address).hton
end
```
Previous output:
```
"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01"
```
New output:
```
"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01"
```

spec/rex/socket_spec.rb
cgranleese-r7: Yea, I have removed that logic now 👍

lib/rex/socket.rb
```diff
@@ -186,14 +186,12 @@ def self.getaddresses(hostname, accept_ipv6 = true)
       return [hostname]
```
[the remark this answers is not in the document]
cgranleese-r7: Got those updated.

lib/rex/socket.rb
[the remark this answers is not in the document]
adfoster-r7: @cgranleese-r7 If I understand correctly, in the scenario of the host being an ipv4 address the logic is short circuited and an array is returned immediately without resolving the host name:
```
require 2.6.5 :001 > require 'rex/socket'
 => true 
2.6.5 :002 > ::Rex::Socket.methods.grep /host/
 => [:gethostbyname] 
2.6.5 :003 > ::Rex::Socket.gethostbyname("127.0.0.1")
 => ["127.0.0.1", [], 2, "\x7F\x00\x00\x01"] 
```

Otherwise we need to call the real method:
```
2.6.5 :004 > ::Rex::Socket.gethostbyname("google.com")
 => ["2a00:1450:4009:81c::200e", "216.58.211.174"] 
```

In this scenario the result is a very different shape, and any calling code will be broken 👀

lib/rex/socket.rb
```diff
@@ -186,14 +186,12 @@ def self.getaddresses(hostname, accept_ipv6 = true)
...
-    res.shift # name
-    res.shift # alias hostnames
-    res.shift # address_family
+    res.map! do |address_info|
+      address_info.ip_address
+    end
+    return [] if not res
 
     # Rubinius has a bug where gethostbyname returns dotted quads instead of
     # NBO, but that's what we want anyway, so just short-circuit here.
```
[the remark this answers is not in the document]
cgranleese-r7: Removed this code as it was tied to the Rubinius bug.

lib/rex/socket.rb
cgranleese-r7: I went for this as before it was shifting the other information out of the array and only returning the IP address. 

I just replicated that, unless you're referring to removing line 191-194 so we return the full address info and not just the IP addresses. 

Whats returned without those lines:
```Ruby
[#<Addrinfo: 127.0.0.1 TCP (localhost)>, #<Addrinfo: ::1 TCP (localhost)>]
```

With those lines:
```Ruby
["127.0.0.1", "::1"]
```

Maybe I'm misunderstanding what you mean though?

cgranleese-r7: > I think these tests need updated similar to how .getaddresses was - rather than deleted 👀

Got those tests added back in and updated.

Gemfile
```diff
@@ -2,3 +2,4 @@ source 'https://rubygems.org'
 
 # Specify your gem's dependencies in rex-socket.gemspec
 gemspec
+gem 'pry', '~> 0.13.1'
```
adfoster-r7, line 5: I don't mind this being here, but it'd be great to move this to being a developer dependency instead if possible :+1:

https://github.com/rapid7/rex-text/blob/fce504e9ea8632f1254e74e83031165ccb723014/Gemfile#L6-L8

lib/rex/socket.rb
```diff
@@ -174,10 +174,9 @@ def self.getaddress(hostname, accept_ipv6 = true)
   end
 
   #
-  # Wrapper for +::Socket.gethostbyname+ that takes special care to see if the
-  # supplied address is already an ASCII IP address.  This is necessary to
-  # prevent blocking while waiting on a DNS reverse lookup when we already
-  # have what we need.
+  # ::Addrinfo.getaddrinfo checks to see if the supplied address is already
+  # an ASCII IP address.  This is necessary to prevent blocking while waiting
+  # on a DNS reverse lookup when we already have what we need.
```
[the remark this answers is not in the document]
cgranleese-r7: Got that added in 👍

lib/rex/socket.rb
```diff
@@ -186,17 +186,12 @@ def self.getaddresses(hostname, accept_ipv6 = true)
...
-    # addresses
-    res.shift # name
-    res.shift # alias hostnames
-    res.shift # address_family
+    res.map! do |address_info|
+      address_info.ip_address
+    end
 
-    # Rubinius has a bug where gethostbyname returns dotted quads instead of
-    # NBO, but that's what we want anyway, so just short-circuit here.
```
adfoster-r7, line 199: https://github.com/rapid7/rex-socket/pull/27#discussion_r577663928

adfoster-r7: Further context to the above comment

The rubinus workaround can be deleted, however the filtering needs to be left in place. I believe it's probably something like this:

```
addr_info = ::Addrinfo.getaddrinfo(hostname, 0, ::Socket::AF_UNSPEC, ::Socket::SOCK_STREAM)
res.map! do |address_info|
    address_info.ip_address
end

unless accept_ipv6
    res.reject! { |ascii| ascii =~ MATCH_IPV6 }
end

res
```

If you wanted to add tests to verify the behavior is still the same, that works for me 🎉

lib/rex/socket.rb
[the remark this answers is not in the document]
adfoster-r7: After discussing with @cgranleese-r7 we might pursue aligning with Ruby's decision to deprecate `gethostbyname` in its entirety.

After looking at metasploit-framework's use of the gethostbyname method, it looks like the common use case is just getting a single address, so an upgrade path might just be something like:

```diff
diff --git a/lib/rex/proto/proxy/socks4a.rb b/lib/rex/proto/proxy/socks4a.rb
index 0c230e4b3d..e3fc73b64b 100644
--- a/lib/rex/proto/proxy/socks4a.rb
+++ b/lib/rex/proto/proxy/socks4a.rb
@@ -132,7 +132,7 @@ class Socks4a
       def resolve( hostname )
         if( not hostname.empty? )
           begin
-            return Rex::Socket.addr_itoa( Rex::Socket.gethostbyname( hostname )[3].unpack( 'N' ).first )
+            return Rex::Socket.getaddress(hostname, false)
           rescue ::SocketError
             return nil
           end
diff --git a/modules/auxiliary/admin/backupexec/dump.rb b/modules/auxiliary/admin/backupexec/dump.rb
index 95ad4c6153..94fe335b88 100644
--- a/modules/auxiliary/admin/backupexec/dump.rb
+++ b/modules/auxiliary/admin/backupexec/dump.rb
@@ -129,7 +129,7 @@ class MetasploitModule < Msf::Auxiliary
       0,
       0,
       1,
-      Rex::Socket.gethostbyname(local_addr)[3],
+      Rex::Socket.resolve_nbo(local_addr, false),
       local_port
     ].pack('NNNNNNNA4N')
diff --git a/modules/auxiliary/dos/windows/smb/ms11_019_electbowser.rb b/modules/auxiliary/dos/windows/smb/ms11_019_electbowser.rb
index 7e1755717d..c5bcd14856 100644
--- a/modules/auxiliary/dos/windows/smb/ms11_019_electbowser.rb
+++ b/modules/auxiliary/dos/windows/smb/ms11_019_electbowser.rb
@@ -56,7 +56,7 @@ class MetasploitModule < Msf::Auxiliary
     @client = Rex::Proto::SMB::Client.new(udp_sock)
     ip = Rex::Socket.source_address(datastore['RHOST'])
-    ip_src = Rex::Socket.gethostbyname(ip)[3]
+    ip_src = Rex::Socket.resolve_nbo(ip, false)
     svc_src = "\x41\x41\x00"   # pre-encoded?
     name_src = Rex::Text.rand_text_alphanumeric(15) # 4+rand(10))
```

There's some internal repositories that make use of `gethostbyname`, so we're on the fence about being good citizens and updating that code too - or leaving that upgrade effort to those teams.
So potential next steps are:

- Leave `::Rex::Socket.gethostbyname` as it is currently, mark it as deprecated, update metasploit-framework to use the new API, and release a future gem that removes it
- Try to update `::Rex::Socket.gethostbyname` to get similar data from `Socket.getaddrinfo`, and add our own deprecation note, and still update metasploit-framework to use the new API. This is probably the most backwards compatible / less impactful change,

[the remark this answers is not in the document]
adfoster-r7: Let's revert this method entirely, mark it as deprecated, and update metasploit framework to no longer use this method :+1:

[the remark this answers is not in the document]
cgranleese-r7: Sounds good, I'll get that sorted 👍

lib/rex/socket.rb
cgranleese-r7: Got these tests added 👍

spec/rex/socket_spec.rb
```diff
@@ -84,52 +84,43 @@
...
-        expect(subject).to eq "1.1.1.1"
-      end
-
-    end
   end
 
   describe '.getaddresses' do
 
-    subject { described_class.getaddresses('whatever') }
+    #TODO: Test to handle IPV6 filtering
```
adfoster-r7, line 113: @cgranleese-r7 This can be removed now

lib/rex/socket.rb
```diff
@@ -217,7 +205,9 @@ def self.getaddresses(hostname, accept_ipv6 = true)
   # not occur.  This is done in order to prevent delays, such as would occur
   # on Windows.
   #
+  # @deprecated as of Ruby 3.0 release.
```
adfoster-r7, line 208:
```suggestion
  # @deprecated Please use {#getaddress}, {#resolv_nbo}, or similar instead.
```

lib/rex/socket.rb
```diff
@@ -186,26 +186,14 @@ def self.getaddresses(hostname, accept_ipv6 = true)
       return [hostname]
     end
 
-    res = ::Socket.gethostbyname(hostname)
-    return [] if not res
+    res = ::Addrinfo.getaddrinfo(hostname, 0, ::Socket::AF_UNSPEC, ::Socket::SOCK_STREAM)
```
adfoster-r7, line 189: Looks like this no longer behaves the same way when it's passed `nil`:
https://github.com/rapid7/rex-sslscan/pull/3#discussion_r712395800

PR #30, opened 2021-03-25
Fix Some RangeWalker issues

zeroSteiner:
In my [previous PR](https://github.com/rapid7/rex-socket/pull/28) as part of refactoring this, I introduced some changes in the behavior of how the `RangeWalker` fails (whoops 😅 ). This corrects those changes to maintain consistency and adds some unit tests so they don't change in the future 🎉 .

This should fix the test failures for rapid7/metasploit-framework#14918.

The main issue is that when multiple things are passed to the RangeWalker, it should gracefully skip the entries that are invalid and keep the ones that are valid. In practice, this meant I needed to switch from returning false to just advancing the loop iteration. I also updated `#expand_nmap` to consistently return `nil` like all of the other parsing functions do when they fail.

## Testing Steps

- [x] Run the new unit tests in isolation, see that they all pass
- [x] Run the unit tests and see that they fix the failures observed in the Metasploit PR<sup>1</sup>

<sup>1</sup>The Metasploit unit tests that are currently failing can be run on their own which is faster using:

```
bundle exec rake db:drop db:create db:migrate RAILS_ENV=test
bundle exec rspec ./spec/models/mdm/workspace_spec.rb
bundle exec rspec ./spec/lib/msf/core/opt_address_range_spec.rb
```

### DIFF
--- a/lib/rex/socket/range_walker.rb
+++ b/lib/rex/socket/range_walker.rb
@@ -84,27 +84,27 @@ class RangeWalker
 
       # Handle IPv6 CIDR first
       if arg.include?(':') && arg.include?('/')
-        return false if (new_ranges = parse_ipv6_cidr(arg)).nil?
+        next if (new_ranges = parse_ipv6_cidr(arg)).nil?
 
       # Handle plain IPv6 next (support ranges, but not CIDR)
       elsif arg.include?(':')
-        return false if (new_ranges = parse_ipv6(arg)).nil?
+        next if (new_ranges = parse_ipv6(arg)).nil?
 
       # Handle IPv4 CIDR
       elsif arg.include?("/")
-        return false if (new_ranges = parse_ipv4_cidr(arg)).nil?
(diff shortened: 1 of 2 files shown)


adfoster-r7:
### Before

Test fail before this change 🔴 

```
Failures:

  1) Rex::Socket::RangeWalker.new should reject an IPv4 address with too many octets
     Failure/Error: ranges += new_ranges
     
     TypeError:
       no implicit conversion of false into Array
     # ./lib/rex/socket/range_walker.rb:110:in `+'
     # ./lib/rex/socket/range_walker.rb:110:in `block in parse'
     # ./lib/rex/socket/range_walker.rb:83:in `each'
     # ./lib/rex/socket/range_walker.rb:83:in `parse'
     # ./lib/rex/socket/range_walker.rb:48:in `initialize'
     # ./spec/rex/socket/range_walker_spec.rb:118:in `new'
     # ./spec/rex/socket/range_walker_spec.rb:118:in `block (3 levels) in <top (require
```

### After

Tests pass after this change 🟢

PR #35, opened 2021-07-19
Fix Default IPv6 LocalHost

### LINKED ISSUE #34: Parameters uses a default IPv4 host for an IPv6 peer
When specifying an IPv6 peer host, the local host value defaults to an IPv4 address.

## Demonstration
```
>> params = Rex::Socket::Parameters.new({'PeerHost' => '::1', 'PeerPort' => 1234})
=> #<Rex::Socket::Parameters:0x000000000bb00a08 @comm=nil, @context=nil, @peerhost="::1", @peerport=1234, @server=nil, @v6=nil>
>> params.localhost
=> "0.0.0.0"
>> 
```

## Proposed Solution
The parameters should intelligently identify when the IPv6 any address should be used instead of the IPv4 value. A basic example would be when only a peer host is specified and that peer host is an IPv6 address.

zeroSteiner:
This updates the `Rex::Socket::Parameters` to intelligently identify when IPv6 should be used while still defaulting to IPv4 when no explicit IPv6 address is specified. This fixes the localhost value to be a proper IPv6 address when it is not specified by the peer as an IPv6 address. The localhost is printed in quite a few locations in Metasploit and can be confusing when it shows something to the effect of `0.0.0.0 -> ::1`.

This also adds quite a few unit tests to ensure that the automatic identification of when to use IPv6 is correct. Of course, explicitly setting the `IPv6` value will supersede the logic. As far as I can tell the local and peer hosts can be one of 4 things, `nil`, a hostname, and IPv4 address or an IPv6 address. A full set of all combinations are included in the unit test and demonstrate when v6 will be set.

This also relies on `#is_ipv4?` and `#is_ipv6?` returning `false` for `nil` as opposed to throwing an exception because the value is not a string. I added a test for this as well to ensure it doesn't change in the future.

Fixes #34

## Verification

- [x] Ensure all of the unit tests pass
- [x] Create parameters to an IPv6 peer, not specifying the local host and see that the localhost is now `::`
- [x] Create parameters to an IPv4 peer, not specifying the local host and see that the localhost is still `0.0.0.0`

### DIFF
--- a/spec/rex/socket/parameters_spec.rb
+++ b/spec/rex/socket/parameters_spec.rb
@@ -15,20 +15,37 @@ RSpec.describe Rex::Socket::Parameters do
 
   describe '.new' do
 
-    it "should handle an IPv4 local host definition" do
-      params = Rex::Socket::Parameters.new({ "LocalHost" => "1.2.3.4", "LocalPort" => 1234 })
-      expect(params.localhost).to eq ("1.2.3.4")
+    it 'should handle an IPv4 local host definition' do
+      params = Rex::Socket::Parameters.new({ 'LocalHost' => '1.2.3.4', 'LocalPort' => 1234 })
+      expect(params.localhost).to eq '1.2.3.4'
       expect(params.localport).to eq 1234
       expect(params.v6?).to eq false
     end
 
-    it "should handle an IPv6 local host definition" do
-      params = Rex::Socket::Parameters.new({ "LocalHost" => "::1", "LocalPort" => 1234, "IPv6" => true })
-      expect(params.localhost).to eq ("::1")
+    it 'should handle an IPv4 peer host definition' do
+      params = Rex::Socket::Parameters.new({ 'PeerHost' => '1.2.3.4', 'PeerPort' => 1234 })
+      expect(params.localhost).to eq '0.0.0.0'
+      expect(params.localport).to eq 0
+      expect(params.peerhost).to eq '1.2.3.4'
+      expect(params.peerport).to eq 1234
+      expect(params.v6?).to eq false
+    end
+
+    it 'should handle an IPv6 local host definition' do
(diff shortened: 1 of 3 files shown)


spec/rex/socket/parameters_spec.rb
```diff
@@ -15,20 +15,74 @@
...
+          expect(params.peerport).to eq 1234
+        end
+
+        it 'should set #v6? correctly' do
+          expect(params.v6?).to eq true
+        end
+      end
+    end
+
+    context "given no value for IPv6" do
```
gwillcox-r7, line 60:
```suggestion
    context 'given no value for IPv6' do
```

gwillcox-r7 (approved):

gwillcox-r7:
Looks like this is working as expected:

```
+ -- --=[ 2154 exploits - 1145 auxiliary - 367 post       ]
+ -- --=[ 592 payloads - 45 encoders - 10 nops            ]
+ -- --=[ 8 evasion                                       ]
Metasploit tip: You can upgrade a shell to a Meterpreter 
session on many platforms using sessions -u 
<session_id>

msf6 > irb
[*] Starting IRB shell...

>> params = Rex::Socket::Parameters.new({'PeerHost' => '::1', 'PeerPort' => 1234})
=> #<Rex::Socket::Parameters:0x00007f43a44f7150 @comm=nil, @context=nil, @peerhost="::1", @peerport=1234, @server=nil, @v6=true>
>> params.localhost
=> "::"
>> params4 = Rex::Socket::Parameters.new({'PeerHost' => '127.0.0.1', 'PeerPort' => 1234})
=> 
#<Rex::Socket::Parameters:0x000055efab6240d0
...
>> params4.localhost
=> "0.0.0.0"
>> params4 = Rex::Socket::Parameters.new({'PeerHost' => '192.168.224.128', 'PeerPort' => 1234})
=> 
#<Rex::Socket::Parameters:0x00007f43a5b14528
...
>> params4.localhost
=> "0.0.0.0"
>> 
```

PR #36, opened 2021-07-19
Use getsockname to get the real local info

### LINKED ISSUE #33: Localhost from getlocalname defaults to the any address
The default value for the localhost defaults to the family specific any-address (`0.0.0.0` for IPv4 and `::` for IPv6). This isn't particularly helpful because it doesn't reflect the actual value used by the socket which is the address of the local interface the socket is bound to. This means when a socket is used in Metasploit, the local address is that any address and not the source IP address of the connection.

This appears to be due to [`#getlocalname`](https://github.com/rapid7/rex-socket/blob/9785327a8cb382ada7555206e73685e57b3d35b3/lib/rex/socket.rb#L733) only calling `#getsockname` when the address is nil and not the any-address.

## Demonstration

```
>> sock4 = Rex::Socket::Tcp.create({'PeerHost' => 'metasploit.com', 'PeerPort' => 443})
=> #<Socket:fd 12>
>> sock4.getlocalname
=> [2, "0.0.0.0", 0]
>> 
>> 
>> sock6 = Rex::Socket::Tcp.create({'PeerHost' => 'google.com', 'PeerPort' => 443, 'IPv6' => true})
=> #<Socket:fd 13>
>> sock6.getlocalname
=> [10, "::", 0]
>> 

>> 
>> sock4.getsockname
=> [2, "192.168.250.129", 38519]
>> sock6.getsockname
=> [10, "::ffff:192.168.250.129", 35159]
>> 
```

## Proposed Solution
`#getlocalname` should be updated to defer to `#getsockname` when the localhost is nil, `0.0.0.0`, or `::`.

zeroSteiner:
This updates `#getlocalname` which caches the address and port to defer to `#getsockname` not only when both values are nil but also when both values are their special any values. This means that the actual IP address the socket is bound to along with it's port number can now be access through the `localhost` and `localport` attributes which provides better information in places where it's printed in Metasploit. This information is typically printed as diagnostic information around connections to show the source.

Fixes #33

## Verification
- [x] Create a local server socket (`server = Rex::Socket::TcpServer.create({'LocalPort' => 0})`)
- [x] Call `#getlocalname` and see a real port number, the IP address will still be `0.0.0.0` by default since it's bound to all interface
- [x] Create a client TCP socket (`sock4 = Rex::Socket::Tcp.create({'PeerHost' => 'metasploit.com', 'PeerPort' => 443})`
- [x] Call `#getlocalname` and see a real, local IP address and port number

## Example
As one example of where this information is printed to the user, see msfconsole's `connect` command:

### Before
```
msf6 exploit(windows/smb/psexec) > connect metasploit.com 80
[*] Connected to metasploit.com:80 (via: 0.0.0.0:0)
```

### After
```
msf6 exploit(windows/smb/psexec) > connect metasploit.com 80
[*] Connected to metasploit.com:80 (via: 192.168.250.87:41417)
```

### DIFF
--- a/lib/rex/socket.rb
+++ b/lib/rex/socket.rb
@@ -730,7 +730,7 @@ module Socket
   # Wrapper around getsockname that stores the local address and local port values.
   #
   def getlocalname
-    if self.localhost.nil? && self.localport.nil?
+    if [nil, '0.0.0.0', '::'].include?(self.localhost) && [nil, 0].include?(self.localport)
       _, self.localhost, self.localport = getsockname
     end
 


gwillcox-r7:
Before:
```
msf6 > irb
[*] Starting IRB shell...

>> sock4 = Rex::Socket::Tcp.create({'PeerHost' => 'metasploit.com', 'PeerPort' => 443})
=> #<Socket:fd 8>
>> sock4.getlocalname
=> [2, "0.0.0.0", 0]
>> sock6 = Rex::Socket::Tcp.create({'PeerHost' => 'google.com', 'PeerPort' => 443, 'IPv6' => true})
=> #<Socket:fd 10>
>> sock6.getlocalname
=> [10, "::", 0]
>> sock4.getsockname
=> [2, "192.168.224.128", 36089]
>> sock6.getsockname
=> [10, "::ffff:192.168.224.128", 41179]
>> 
```

After:
```
msf6 > connect metasploit.com 80
[*] Connected to metasploit.com:80 (via: 192.168.224.128:33255)
^Cmsf6 > irb
[*] Starting IRB shell...

>> sock4 = Rex::Socket::Tcp.create({'PeerHost' => 'metasploit.com', 'PeerPort' => 443})
=> #<Socket:fd 8>
>> sock4.getlocalname
=> [2, "192.168.224.128", 41059]
>> sock6 = Rex::Socket::Tcp.create({'PeerHost' => 'google.com', 'PeerPort' => 443, 'IPv6' => true})
=> #<Socket:fd 10>
>> sock6.getlocalname
=> [10, "::ffff:192.168.224.128", 34605]
>> sock6.getsockname
=> [10, "::ffff:192.168.224.128", 34605]
>> sock4.getsockname
=> [2, "192.168.224.128", 41059]
>> 
```

Works as intended, will land this now :)

PR #40, opened 2021-09-24
Set underlying comm value during initialization and do not change it

smashery:
This change supports [PR 15706](https://github.com/rapid7/metasploit-framework/pull/15706) in Metasploit.

The fact that `comm` could change at any time (i.e. as the route table changes) made it very difficult to reason about the state of the parameters object. If the route table is changed after a session is established, then the params object associated with the session object could then change its `comm` value. It is helpful for a Metasploit session object to use the params object to initialize sockets; but if they could change between creation of the session and creation of the socket, this introduces a race condition whereby a session is appropriately set up, but a socket it creates (e.g. via TCP forwarding) believes it's being passed through a different session, or none at all.

This commit also adds the `comm` attribute to sockets, so that sockets can deduce which session they're listening on.

### DIFF
--- a/lib/rex/socket.rb
+++ b/lib/rex/socket.rb
@@ -707,6 +707,7 @@ module Socket
       self.peerport  = params.peerport
       self.localhost = params.localhost
       self.localport = params.localport
+      self.comm      = params.comm
       self.context   = params.context || {}
       self.ipv       = params.v6 ? 6 : 4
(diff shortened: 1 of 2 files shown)


smashery:
Actually never mind; I found the API I should have been using instead; this PR won't be needed for the other change.

PR #43, opened 2022-03-29
Prevent DNS leaks & pivot using optional resolver

sempervictus:
`Rex::Socket` eventually ends up relying on the standard library
for name resolution in socket-related calls. Specifically, the use
of `Socket.gethostbyname` and `Addrinfo.getaddrinfo` can lead to
the use of system DNS, recursing into places where users may not
want to send DNS requests.

Implement a mechanism to set a global name resolver for all of
`Rex::Socket` producing `.get('msf.io')` output analogous to that
of `Rex::Proto::DNS::Resolver`. Once that resolver is in-place,
the class method calls will be diverted to internal ones using
said resolver to create API-compatible output structures
feeding transparently back to the caller. Unlike approaches using
`Rex::Socket::SwitchBoard` directly, this mechanism can't create
loops in the stack as the `Resolver` uses IP-based NS' and cannot
get stuck looking up the name servers against which to look-up...

Clean up raw calls to `Socket` and ensure that all `gethostbyname`
calls are properly wrapped.

Note: a fully-compatible resolver would cache DNS requests until
the TTL on them expires and honor the local system hostsfile just
like the native standard library does. However, that would fall
into `rex/proto` or other parts of the ecosystem and not in-scope
for this repository. As a result, the operating semantic of this
library is unchanged by default and requires input of a relevant
`Resolver` from other components which are capable of constructing
and configuring one as appropriate.

Resolver tasks:
Optimize for performance and concurrent access (w/ Rex::Proto)
Review options for per-caller instantiation or NS passing by the
caller to permit dynamic dispatch to pivoted nameservers past the
`SwitchBoard` - this is likely to be handled by the mechanism which
will deal with the hostsfile concern such as to use session and
workspace data to determine which NS to query for an FQDN and how.

### DIFF
--- a/lib/rex/socket.rb
+++ b/lib/rex/socket.rb
@@ -893,6 +893,9 @@ protected
     if !v4 and !v6
       raise ::SocketError.new('getaddrinfo: Name or service not known')
     end
+    [v4, v6].compact.map do |ans|
+      ans = resolver.send(ans.name).answer.first unless ans.respond_to?(:address)
+    end
     # Build response array
     getaddrinfo = []
     getaddrinfo << Addrinfo.new(
--- a/lib/rex/socket.rb
+++ b/lib/rex/socket.rb
@@ -918,12 +918,20 @@ protected
       "Rex::Socket internal DNS resolution requires passing a String name to resolve"
     ) unless name.is_a?(String)
     # Pull both record types
-    v4 = resolver.send(name, ::Net::DNS::A).answer.select {
-      |a| a.type == Dnsruby::Types::A}.sort_by {|a| a.address.to_s
-    }
-    v6 = resolver.send(name, ::Net::DNS::AAAA).answer.select {|a|
-      a.type == Dnsruby::Types::AAAA}.sort_by {|a| a.address.to_s
-    }
+    v4 = begin
+      resolver.send(name, ::Net::DNS::A).answer.select {|a|
+        a.type == Dnsruby::Types::A}.sort_by {|a| a.address.to_s
+      }
+    rescue
+      [nil]
+    end
+    v6 = begin
+      resolver.send(name, ::Net::DNS::AAAA).answer.select {|a|
+        a.type == Dnsruby::Types::AAAA}.sort_by {|a| a.address.to_s
+      }
+    rescue
+      [nil]
+    end
     # Emulate ::Socket's error if no responses found
     if !v4[0] and !v6[0]
       raise ::SocketError.new('getaddrinfo: Name or service not known')
--- a/lib/rex/socket.rb
+++ b/lib/rex/socket.rb
@@ -729,6 +729,8 @@ module Socket
   #
   # Install Rex::Proto::DNS::CachedResolver, or similar, to pivot DNS
   #
+  # @param res [Rex::Proto::DNS::CachedResolver] Resolver object to handle DNS requests
+  # @return [Rex::Proto::DNS::CachedResolver] The installed resolver
   def self._install_global_resolver(res)
     @@resolver = res
   end
@@ -858,17 +860,12 @@ protected
   attr_writer :context # :nodoc:
   attr_writer :ipv # :nodoc:
 
+  #
+  # @param name [String] The hostname to lookup via the resolver
+  # @param resolver [Rex::Proto::DNS::CachedResolver] Resolver to query for the name
+  # @return [Array] Array mimicking the native gethostbyname return type
   def self.rex_gethostbyname(name, resolver = @@resolver)
-    raise ::SocketError.new(
-      "Rex::Socket internal DNS resolution requires passing/setting a resolver"
-    ) unless resolver
-    # Pull both record types
-    v4 = resolver.send(name, ::Net::DNS::A).answer.first
-    v6 = resolver.send(name, ::Net::DNS::AAAA).answer.first
-    # Emulate ::Socket's error if no responses found
-    if !v4 and !v6
-      raise ::SocketError.new('getaddrinfo: Name or service not known')
-    end
+    v4, v6 = self.rex_resolve_hostname(name, resolver)
     # Build response array
     hostbyname = [name, []]
     if v4
@@ -882,20 +879,12 @@ protected
     return hostbyname
   end
 
+  #
+  # @param name [String] The hostname to lookup via the resolver
+  # @param resolver [Rex::Proto::DNS::CachedResolver] Resolver to query for the name
+  # @return [Array] Array mimicking the native getaddrinfo return type
   def self.rex_getaddrinfo(name, resolver = @@resolver)
-    raise ::SocketError.new(
-      "Rex::Socket internal DNS resolution requires passing/setting a resolver"
-    ) unless resolver
-    # Pull both record types
-    v4 = resolver.send(name, ::Net::DNS::A).answer.first
-    v6 = resolver.send(name, ::Net::DNS::AAAA).answer.first
-    # Emulate ::Socket's error if no responses found
-    if !v4 and !v6
-      raise ::SocketError.new('getaddrinfo: Name or service not known')
-    end
-    [v4, v6].compact.map do |ans|
-      ans = resolver.send(ans.name).answer.first unless ans.respond_to?(:address)
-    end
+    v4, v6 = self.rex_resolve_hostname(name, resolver)
     # Build response array
     getaddrinfo = []
     getaddrinfo << Addrinfo.new(
@@ -911,6 +900,31 @@ protected
       ::Socket::IPPROTO_TCP,
     ) if v6
     return getaddrinfo
+  end
+
+
+  # @param name [String] The hostname to lookup via the resolver
+  # @param resolver [Rex::Proto::DNS::CachedResolver] Resolver to query for the name
+  # @return [Array] Array of Dnsruby::Message responses for consumers to reformat
+  def self.rex_resolve_hostname(name, resolver = @@resolver)
+    raise ::SocketError.new(
+      "Rex::Socket internal DNS resolution requires passing/setting a resolver"
+    ) unless resolver
+    raise ::SocketError.new(
+      "Rex::Socket internal DNS resolution requires passing a String name to resolve"
+    ) unless name.is_a?(String)
+    # Pull both record types
+    v4 = resolver.send(name, ::Net::DNS::A).answer.first
+    v6 = resolver.send(name, ::Net::DNS::AAAA).answer.first
+    # Emulate ::Socket's error if no responses found
+    if !v4 and !v6
+      raise ::SocketError.new('getaddrinfo: Name or service not known')
+    end
+    # Ensure response types (depending on underlying library used) provide required methods
+    [v4, v6].compact.map do |ans|
+      ans = resolver.send(ans.name).answer.first unless ans.respond_to?(:address)
+    end
+    return v4, v6
   end
 end
 
--- a/lib/rex/socket.rb
+++ b/lib/rex/socket.rb
@@ -855,48 +855,54 @@ protected
   attr_writer :context # :nodoc:
   attr_writer :ipv # :nodoc:
 
-  def self.rex_gethostbyname(name)
+  def self.rex_gethostbyname(name, resolver = @@resolver)
+    raise ::SocketError.new(
+      "Rex::Socket internal DNS resolution requires passing/setting a resolver"
+    ) unless resolver
     # Pull both record types
-    v4 = @@resolver.send(name, ::Net::DNS::A).answer.first
-    v6 = @@resolver.send(name, ::Net::DNS::AAAA).answer.first
+    v4 = resolver.send(name, ::Net::DNS::A).answer.first
+    v6 = resolver.send(name, ::Net::DNS::AAAA).answer.first
     # Emulate ::Socket's error if no responses found
     if !v4 and !v6
-      raise SocketError.new('getaddrinfo: Name or service not known')
+      raise ::SocketError.new('getaddrinfo: Name or service not known')
     end
     # Build response array
     hostbyname = [name, []]
     if v4
-      hostbyname << Socket::AF_INET
+      hostbyname << ::Socket::AF_INET
       hostbyname << self.addr_aton(v4.address.to_s)
       hostbyname << self.addr_aton(v6.address.to_s) if v6
     else
-      hostbyname << Socket::AF_INET6
+      hostbyname << ::Socket::AF_INET6
       hostbyname << self.addr_aton(v6.address.to_s)
     end
     return hostbyname
   end
 
-  def self.rex_getaddrinfo(name)
+  def self.rex_getaddrinfo(name, resolver = @@resolver)
+    raise ::SocketError.new(
+      "Rex::Socket internal DNS resolution requires passing/setting a resolver"
+    ) unless resolver
     # Pull both record types
-    v4 = @@resolver.send(name, ::Net::DNS::A).answer.first
-    v6 = @@resolver.send(name, ::Net::DNS::AAAA).answer.first
+    v4 = resolver.send(name, ::Net::DNS::A).answer.first
+    v6 = resolver.send(name, ::Net::DNS::AAAA).answer.first
     # Emulate ::Socket's error if no responses found
     if !v4 and !v6
-      raise SocketError.new('getaddrinfo: Name or service not known')
+      raise ::SocketError.new('getaddrinfo: Name or service not known')
     end
     # Build response array
     getaddrinfo = []
     getaddrinfo << Addrinfo.new(
       self.to_sockaddr(v4.address.to_s,0),
-      Socket::AF_INET,
-      Socket::SOCK_STREAM,
-      Socket::IPPROTO_TCP,
+      ::Socket::AF_INET,
+      ::Socket::SOCK_STREAM,
+      ::Socket::IPPROTO_TCP,
     ) if v4
     getaddrinfo << Addrinfo.new(
       self.to_sockaddr(v6.address.to_s,0),
-      Socket::AF_INET6,
-      Socket::SOCK_STREAM,
-      Socket::IPPROTO_TCP,
+      ::Socket::AF_INET6,
+      ::Socket::SOCK_STREAM,
+      ::Socket::IPPROTO_TCP,
     ) if v6
     return getaddrinfo
   end
--- a/lib/rex/socket.rb
+++ b/lib/rex/socket.rb
@@ -216,7 +216,11 @@ module Socket
       return [hostname]
     end
 
-    res = ::Addrinfo.getaddrinfo(hostname, 0, ::Socket::AF_UNSPEC, ::Socket::SOCK_STREAM)
+    res = if @@resolver
+      self.rex_getaddrinfo(hostname)
+    else
+      ::Addrinfo.getaddrinfo(hostname, 0, ::Socket::AF_UNSPEC, ::Socket::SOCK_STREAM)
+    end
 
     res.map! do |address_info|
       address_info.ip_address
@@ -248,7 +252,7 @@ module Socket
       host, _ = host.split('%', 2)
     end
 
-    ::Socket.gethostbyname(host)
+    @@resolver ? self.rex_gethostbyname(host) : ::Socket.gethostbyname(host)
   end
 
   #
@@ -719,6 +723,13 @@ module Socket
     return [lsock, rsock]
   end
 
+  #
+  # Install Rex::Proto::DNS::Resolver, or similar, to pivot DNS
+  #
+  def self._install_global_resolver(res)
+    @@resolver = res
+  end
+
 
   ##
   #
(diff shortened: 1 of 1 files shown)


timwr:
@sempervictus can you provide verification steps for this please?

gwillcox-r7:
@sempervictus Any update on testing directions for this?

sempervictus:
@gwillcox-r7 - thanks for the ping. Here's what i use for my own configuration, including enabling my framework to host reverse sessions on TOR:
```ruby
  require 'rex/proto/dns/resolver'
  resolver = Rex::Proto::DNS::CachedResolver.new
  resolver.nameservers = %w{8.8.4.4} # or whatever you want
  Rex::Socket._install_global_resolver resolver
  # Hack-in self-tor
  Rex::Socket.class_variable_get(:@@resolver).cache.add_static('mylongaddressstring.onion','127.0.0.1')
  ```

**Comment by sempervictus** [comment]
Should be more or less solid as of that last commit.
Testing setup script:
```ruby
  require 'rex/proto/dns/resolver'
  resolver = Rex::Proto::DNS::CachedResolver.new
  resolver.proxies = 'socks4:127.0.0.1:9050' if ENV['DNS_OVER_TOR']
  resolver.nameservers = %w{8.8.4.4} # or whatever you want
  Rex::Socket._install_global_resolver resolver
```
Then sniff your DNS queries and verify that you are not hitting the OS' resolver and that  `proxies` work as intended. If you want to play on `Ultra Violence`, you can introspect on the `CachedResolver` and its `Cache`, and if you play on `Nightmare` then try overriding IP resolution for all of framework by using `resolver.cache.add_static('hostname','x.x.x.x')`

gwillcox-r7:
Given this hasn't seen movement in nearly a year I'm reassigning this to myself so we can move this forwards.

lib/rex/socket.rb
```diff
@@ -190,7 +193,11 @@ def self.getaddresses(hostname, accept_ipv6 = true)
       return [hostname]
     end
 
-    res = ::Addrinfo.getaddrinfo(hostname, 0, ::Socket::AF_UNSPEC, ::Socket::SOCK_STREAM)
+    res = if @@resolver
+      self.rex_getaddrinfo(hostname)
```
gwillcox-r7, line 197: Should you not also be passing in the other information here???

lib/rex/socket.rb
```diff
@@ -693,6 +700,13 @@ def self.udp_socket_pair
     return [lsock, rsock]
   end
 
+  #
+  # Install Rex::Proto::DNS::CachedResolver, or similar, to pivot DNS
+  #
+  def self._install_global_resolver(res)
```
gwillcox-r7, line 706: Would be good to add some YARD documentation on the `res` parameter here.

lib/rex/socket.rb
```diff
@@ -809,10 +828,64 @@ def type?
 
 protected
 
-  attr_writer :peerhost, :peerport, :localhost, :localport # :nodoc:
+  attr_writer :peerhost, :peerhostname, :peerport, :localhost, :localport # :nodoc:
   attr_writer :context # :nodoc:
   attr_writer :ipv # :nodoc:
 
+  def self.rex_gethostbyname(name, resolver = @@resolver)
```
gwillcox-r7, line 835: Missing YARD documentation here, which would be quite helpful, particularly for understanding what these parameters types are meant to be.

lib/rex/socket.rb
```diff
@@ -809,10 +828,64 @@ def type?
...
+      hostbyname << self.addr_aton(v4.address.to_s)
+      hostbyname << self.addr_aton(v6.address.to_s) if v6
+    else
+      hostbyname << ::Socket::AF_INET6
+      hostbyname << self.addr_aton(v6.address.to_s)
+    end
+    return hostbyname
+  end
+
+  def self.rex_getaddrinfo(name, resolver = @@resolver)
```
gwillcox-r7, line 859: Missing YARD documentation here, which would be quite helpful, particularly for understanding what these parameters types are meant to be.

lib/rex/socket.rb
```diff
@@ -809,10 +828,64 @@ def type?
...
   attr_writer :context # :nodoc:
   attr_writer :ipv # :nodoc:
 
+  def self.rex_gethostbyname(name, resolver = @@resolver)
+    raise ::SocketError.new(
+      "Rex::Socket internal DNS resolution requires passing/setting a resolver"
+    ) unless resolver
+    # Pull both record types
+    v4 = resolver.send(name, ::Net::DNS::A).answer.first
+    v6 = resolver.send(name, ::Net::DNS::AAAA).answer.first
```
gwillcox-r7, line 841: Forgive me but this seems a little suspicious. What happens if the resolve doesn't respond at all? If it returns an empty  array in that case then this is fine but otherwise if it returns `nil` then this will break.

This also applies to the same pattern in your other function below.

lib/rex/socket.rb
```diff
@@ -809,10 +828,64 @@ def type?
...
+    # Emulate ::Socket's error if no responses found
+    if !v4 and !v6
+      raise ::SocketError.new('getaddrinfo: Name or service not known')
+    end
+    # Build response array
+    hostbyname = [name, []]
+    if v4
+      hostbyname << ::Socket::AF_INET
+      hostbyname << self.addr_aton(v4.address.to_s)
+      hostbyname << self.addr_aton(v6.address.to_s) if v6
```
gwillcox-r7, line 851: This assumes that the `address` fields are correctly populated from earlier. Might be good to check their contents prior to casting like this.

lib/rex/socket.rb
```diff
@@ -809,10 +828,64 @@ def type?
...
+    ) unless resolver
+    # Pull both record types
+    v4 = resolver.send(name, ::Net::DNS::A).answer.first
+    v6 = resolver.send(name, ::Net::DNS::AAAA).answer.first
+    # Emulate ::Socket's error if no responses found
+    if !v4 and !v6
+      raise ::SocketError.new('getaddrinfo: Name or service not known')
+    end
+    [v4, v6].compact.map do |ans|
+      ans = resolver.send(ans.name).answer.first unless ans.respond_to?(:address)
```
gwillcox-r7, line 871: Same concern here over if there is no answer, what the value of `answer` will be.

gwillcox-r7:
> ```ruby
>  require 'rex/proto/dns/resolver'
>   resolver = Rex::Proto::DNS::CachedResolver.new
> ```

When I was testing this the require line was failing to import and I can't find any reference to `Rex::Proto::DNS::CachedResolver` anywhere. Is that a custom implementation you have made?

lib/rex/socket.rb
sempervictus: I think this pattern comes from the old net/dns code on which i started hacking the DNS horror show way back.
I'm all ears on proposed improvements

lib/rex/socket.rb
sempervictus: Do we want to type-check with `is_a?` or the like?

lib/rex/socket.rb
sempervictus: if you compact the array and it contains `nil`s then the `do` block will never run, i think :).

lib/rex/socket.rb
gwillcox-r7: Ah the joys of old code. I'd probably break this and similar lines down so instead of doing everything on one line we do something like:
1. send
2. Check the response isn't NIL
3. Check the response has an answer with at least one element.
4. Call .first to get the first element.

lib/rex/socket.rb
gwillcox-r7: Yep pretty much. Just wanted to check that it wasn't blank and we are trying to act on a blank response. So something like `.blank?` and a check for the type here would be good?

lib/rex/socket.rb
gwillcox-r7: True true.

Also now that I look at it, do we need this assignment to `ans` here? It will be updated on every run of the loop so the assignment result will just be discarded?

sempervictus: IIRC its a compatibility hack of some sort to make callers of the return happy. Unfortunately i'm pretty sure my memory works kind of like an ARC cache and this data slipped MFU and LRU sides of the cache :smile: - will try to divine why i would do something so inefficient.

lib/rex/socket.rb
sempervictus: Ah, ok, so this is actually "safe code" - the `Rex::Proto::DNS::CachedResolver`  can _only_  produce a `Dnsruby::Message` when `:send`ing any `String` to the resolver (`pry(Rex::Socket)> v4 = resolver.send('', ::Net::DNS::A).class => Dnsruby::Message`) - so we actually need to verify _name_ not the rest of it for "type safety:"
```ruby
[7] pry(Rex::Socket)> name = 'somerandomthingthatwontresolve.nil'
=> "somerandomthingthatwontresolve.nil"
[8] pry(Rex::Socket)> v4 = resolver.send(name, ::Net::DNS::A).answer.first
=> nil
[9] pry(Rex::Socket)> v4 = resolver.send(name, ::Net::DNS::A).answer
=> []
[10] pry(Rex::Socket)> v4 = resolver.send(name, ::Net::DNS::A).class
=> Dnsruby::Message
```
so thinking
```ruby
  #
  # @param name [String] The hostname to lookup via the resolver
  # @param resolver [Rex::Proto::DNS::CachedResolver] Resolver to query for the name
  # @return [Array] Array mimicking the native getaddrinfo return type
  def self.rex_getaddrinfo(name, resolver = @@resolver)
    raise ::SocketError.new(
      "Rex::Socket internal DNS resolution requires passing/setting a resolver"
    ) unless resolver
    raise ::SocketError.new(
      "Rex::Socket internal DNS resolution requires passing/setting String name to resolve"
    ) unless name.is_a?(String)
```
should do it

sempervictus:
Eh, so that merge conflict is from **exactly** the sort of thing i'm trying to fix here - we _cannot_ have `::Socket` calls for name resolution anywhere in Rex above the code handled by this PR or it will leak DNS:
```diff
435     when 'socks4'
436 <<<<<<< HEAD
437       supports_ipv6 = false
438       setup = [4,1,port.to_i].pack('CCn') + Socket.resolv_nbo(host, supports_ipv6) + Rex::Text.rand_text_alpha(rand(8)+1) + "\x00"
439 =======
440       setup = [4,1,port.to_i].pack('CCn') + Rex::Socket.gethostbyname(host)[3] + Rex::Text.rand_text_alpha(rand(8)+1) + "\x00"
441 >>>>>>> 6d862ea (Prevent DNS leaks & pivot using optional resolver)
```
The intent of this effort is not only to provide a safe and pivoting resolver mechanism, but also a stable API for consumers to call into.
I'm going to rebase and fix, but once this is in, we should probably audit all consumers for this sort of thing and adjust to use `Rex::Socket` APIs

lib/rex/socket.rb
gwillcox-r7: This wasn't marked as resolved but confirmed that the updates fix this issue so marking as resolved.

lib/rex/socket.rb
```diff
@@ -818,6 +834,72 @@ def type?
...
+  #
+  # @param name [String] The hostname to lookup via the resolver
+  # @param resolver [Rex::Proto::DNS::CachedResolver] Resolver to query for the name
+  # @return [Array] Array mimicking the native getaddrinfo return type
+  def self.rex_getaddrinfo(name, resolver = @@resolver)
+    v4, v6 = self.rex_resolve_hostname(name, resolver)
+    # Build response array
+    getaddrinfo = []
+    getaddrinfo << Addrinfo.new(
+      self.to_sockaddr(v4.address.to_s,0),
```
gwillcox-r7, line 865: Same point as above here r.e `address` value not being checked.

gwillcox-r7:
Unresolving the YARD documentation since `Rex::Proto::DNS::CachedResolver` isn't a class that exists anywhere. I appreciate you trying to make the name clearer but if it isn't a class that exists then your going to make tools like Solargraph hellya confused when they try to use that to do stuff like Go To Definition only to find that the class doesn't exist.

lib/rex/socket.rb
```diff
@@ -222,7 +229,7 @@ def self.gethostbyname(host)
       host, _ = host.split('%', 2)
     end
 
-    ::Socket.gethostbyname(host)
+    @@resolver ? self.rex_gethostbyname(host) : ::Socket.gethostbyname(host)
```
gwillcox-r7, line 232: If either of these calls fail, we should probably be handling this appropriately vs blindly assigning to @@resolver, particularly given both of these will require reaching out to remote hosts to perform actions.

sempervictus: What do you mean? Only one of those calls (`self.rex_gethostbyname(host)`) can occur when `@@resolver` is defined, otherwise it will use the underlying `Socket` method

sempervictus:
> Unresolving the YARD documentation since `Rex::Proto::DNS::CachedResolver` isn't a class that exists anywhere. I appreciate you trying to make the name clearer but if it isn't a class that exists then your going to make tools like Solargraph hellya confused when they try to use that to do stuff like Go To Definition only to find that the class doesn't exist.

The class exists, but outside the scope of _this_ code since its passed down from Metasploit (maybe `rex-proto` should be broken out into gems?).
```ruby
pry(#<Msf::Framework>)> show-source Rex::Proto::DNS::CachedResolver

From: /opt/metasploit4/msf4/lib/rex/proto/dns/resolver.rb:374
Class name: Rex::Proto::DNS::CachedResolver
Number of lines: 87

class CachedResolver < Resolver
  include Rex::Proto::DNS::Constants
  attr_accessor :cache
...
```

gwillcox-r7:
@sempervictus I presume you are looking at https://github.com/rapid7/metasploit-framework/blob/master/lib/rex/proto/dns/resolver.rb? I don't see any `CachedResolver` there. I also tried looking across all Rapid7 repos with https://github.com/search?q=org%3Arapid7+CachedResolver&type=code but didn't find anything, which is why I'm curious if this is perhaps local code that you had given you tend to do customizations to our code?

Forgive me if I'm misunderstanding something though.

sempervictus:
Oh humbug, it died on the vine in the PR queue - https://github.com/rapid7/metasploit-framework/pull/16390.
Will see what that needs and try to revive it.

sempervictus:
The only delta i have in that file w/ my fork is lack of some of the YARD annotations in _my_ version - looks like what's in the PR is actually newer than mine.

smcintyre-r7:
I'd be alot more confident in moving this forward if someone could test the `auxiliary/gather/enum_dns` with these changes applied. Additionally any other module using the `Msf::Exploit::Remote::DNS::Enumeration` mixin.

Also if we're preventing DNS leaks then I'm assuming (and someone please correct me if I'm wrong) that this would be impacting all of the DNS queries we're making through rex and thus the RangeWalker should also be thoroughly inspected to ensure it's still going to work as intended.

lib/rex/socket.rb
gwillcox-r7: I honestly don't know why I was thinking assignment was involved here. I blame looking through this when I was too tired 😓 Sorry for the confusion, marking this as resolved.

lib/rex/socket.rb
gwillcox-r7: Since been resolved.

lib/rex/socket.rb
```diff
@@ -190,7 +193,11 @@ def self.getaddresses(hostname, accept_ipv6 = true)
       return [hostname]
     end
 
-    res = ::Addrinfo.getaddrinfo(hostname, 0, ::Socket::AF_UNSPEC, ::Socket::SOCK_STREAM)
+    res = if @@resolver
+      self.rex_getaddrinfo(hostname)
+    else
+      ::Addrinfo.getaddrinfo(hostname, 0, ::Socket::AF_UNSPEC, ::Socket::SOCK_STREAM)
+    end
```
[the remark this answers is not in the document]
gwillcox-r7: Resolved with https://github.com/rapid7/rex-socket/pull/43/commits/21597c92276ea560b05c7cbbfea20df05788ed06

lib/rex/socket.rb
gwillcox-r7: This is negated in part by this code:

```
    # Emulate ::Socket's error if no responses found
    if !v4[0] and !v6[0]
      raise ::SocketError.new('getaddrinfo: Name or service not known')
    end
```

This is in `  def self.rex_resolve_hostname(name, resolver = @@resolver)` inside `lib/rex/socket.rb`.

lib/rex/socket.rb
gwillcox-r7: The inefficient code has been removed and the new code looks good so going to mark this as resolved.

sempervictus:
Rangewalkers work far as i know - i use this code every day, though i would appreciate another pair of eyes making sure that the PRs cover all uses of the methods stubbed out in Rex for address acquisition.
Will take a peek at @gwillcox-r7' changes this evening (thanks)

lib/rex/socket.rb
```diff
@@ -193,10 +193,10 @@ def self.getaddresses(hostname, accept_ipv6 = true)
       return [hostname]
     end
 
-    res = if @@resolver
-      self.rex_getaddrinfo(hostname)
+    if @@resolver
+      res = self.rex_getaddrinfo(hostname)
     else
-      ::Addrinfo.getaddrinfo(hostname, 0, ::Socket::AF_UNSPEC, ::Socket::SOCK_STREAM)
+      res = ::Addrinfo.getaddrinfo(hostname, 0, ::Socket::AF_UNSPEC, ::Socket::SOCK_STREAM)
```
sempervictus, line 199: Sorry, Rust-y patterns used to get something owned by context outside the conditional leaking into my Ruby code. Appreciate the fix.

lib/rex/socket.rb
gwillcox-r7: > Ah, ok, so this is actually "safe code" - the `Rex::Proto::DNS::CachedResolver` can _only_ produce a `Dnsruby::Message` when `:send`ing any `String` to the resolver (`pry(Rex::Socket)> v4 = resolver.send('', ::Net::DNS::A).class => Dnsruby::Message`) - so we actually need to verify _name_ not the rest of it for "type safety:"
> 
> ```ruby
> [7] pry(Rex::Socket)> name = 'somerandomthingthatwontresolve.nil'
> => "somerandomthingthatwontresolve.nil"
> [8] pry(Rex::Socket)> v4 = resolver.send(name, ::Net::DNS::A).answer.first
> => nil
> [9] pry(Rex::Socket)> v4 = resolver.send(name, ::Net::DNS::A).answer
> => []
> [10] pry(Rex::Socket)> v4 = resolver.send(name, ::Net::DNS::A).class
> => Dnsruby::Message
> ```
> 
> so thinking
> 
> ```ruby
>   #
>   # @param name [String] The hostname to lookup via the resolver
>   # @param resolver [Rex::Proto::DNS::CachedResolver] Resolver to query for the name
>   # @return [Array] Array mimicking the native getaddrinfo return type
>   def self.rex_getaddrinfo(name, resolver = @@resolver)
>     raise ::SocketError.new(
>       "Rex::Socket internal DNS resolution requires passing/setting a resolver"
>     ) unless resolver
>     raise ::SocketError.new(
>       "Rex::Socket internal DNS resolution requires passing/setting String name to resolve"
>     ) unless name.is_a?(String)
> ```
> 
> should do it

If I'm not mistaken the code at https://github.com/rapid7/metasploit-framework/pull/16390/files#diff-0e9da515e448f98bae7ee7685586d474b02635232727f8265a0ef1b931281d04R140 shows this could also raise an exception or return `Nil`? So this would actually fail if it returned `nil`.

sempervictus: That's fine - https://github.com/rapid7/rex-socket/pull/43/files#diff-4413927154bdecf1370ea1e11874d39c9760d8b8174014ec8dca4dd0798b4e96R901 handles that situation and appropriately raises.

sempervictus:
OK, glanced over this in a bit more detail, sorry for lag.
@smcintyre-r7 - that is _very_ much how i use this code (`dns_enum`), except that i've implemented a bunch of BRE around the queries to deal with parallel resolution over TOR timing out/failing:
```diff
diff --git c/lib/msf/core/exploit/remote/dns/enumeration.rb w/lib/msf/core/exploit/remote/dns/enumeration.rb
index 3ffe7e8d98..0eb783ca56 100644
--- c/lib/msf/core/exploit/remote/dns/enumeration.rb
+++ w/lib/msf/core/exploit/remote/dns/enumeration.rb
@@ -115,7 +115,11 @@ module Enumeration
   end
 
   def dns_get_a(domain, type='DNS A records', displayed=false)
-    resp = dns_query(domain, 'A')
+    begin
+      resp = dns_query(domain, 'A')
+    rescue NoResponseError
+      return
+    end
     return if resp.blank? || resp.answer.blank?
 
     records = []
@@ -131,7 +135,11 @@ module Enumeration
 
   def dns_get_cname(domain)
     print_status("Querying DNS CNAME records for #{domain}")
-    resp = dns_query(domain, 'CNAME')
+    begin
+      resp = dns_query(domain, 'CNAME')
+    rescue NoResponseError
+      return
+    end
     return if resp.blank? || resp.answer.blank?
 
     records = []
@@ -157,7 +165,7 @@ module Enumeration
         records << r.exchange.to_s
         print_good("#{domain} MX: #{r.exchange}")
       end
-    rescue SocketError => e
+    rescue NoResponseError, SocketError => e
       print_error("Query #{domain} DNS MX - exception: #{e}")
     ensure
       return if records.blank?
@@ -168,7 +176,11 @@ module Enumeration
 
   def dns_get_ns(domain)
     print_status("Querying DNS NS records for #{domain}")
-    resp = dns_query(domain, 'NS')
+    begin
+      resp = dns_query(domain, 'NS')
+    rescue NoResponseError
+      return
+    end
     return if resp.blank? || resp.answer.blank?
 
     records = []
@@ -183,7 +195,11 @@ module Enumeration
   end
 
   def dns_get_ptr(ip)
-    resp = dns_query(ip, nil)
+    begin
+      resp = dns_query(ip)
+    rescue NoResponseError
+      return
+    end
     return if resp.blank? || resp.answer.blank?
 
     records = []
@@ -199,7 +215,11 @@ module Enumeration
 
   def dns_get_soa(domain)
     print_status("Querying DNS SOA records for #{domain}")
-    resp = dns_query(domain, 'SOA')
+    begin
+      resp = dns_query(domain, 'SOA')
+    rescue NoResponseError
+      return
+    end
     return if resp.blank? || resp.answer.blank?
 
     records = []
@@ -227,7 +247,11 @@ module Enumeration
     srv_record_types.each do |srv_record_type|
       srv_protos.each do |srv_proto|
         srv_record = "_#{srv_record_type}._#{srv_proto}.#{domain}"
-        resp = dns_query(srv_record, Net::DNS::SRV)
+        begin
+          resp = dns_query(srv_record, Net::DNS::SRV)
+        rescue NoResponseError
+          next
+        end
         next if resp.blank? || resp.answer.blank?
         srv_record_data = []
         resp.answer.each do |r|
@@ -305,7 +329,11 @@ module Enumeration
 
   def dns_get_txt(domain)
     print_status("Querying DNS TXT records for #{domain}")
-    resp = dns_query(domain, 'TXT')
+    begin
+      resp = dns_query(domain, 'TXT')
+    rescue NoResponseError
+      return
+    end
     return if resp.blank? || resp.answer.blank?
 
     records = []
```

lib/rex/socket.rb
gwillcox-r7: I thought `resolver.send(name, ::Net::DNS::A)` would return and then we would call `.answer.select` on a `nil` object which would cause errors?

gwillcox-r7: Resolved in https://github.com/rapid7/rex-socket/pull/43/files/21597c92276ea560b05c7cbbfea20df05788ed06..52881ad9232840fe15111b24ac60e9e2e1d2e1d8

sempervictus:
In case it eases merge concerns: simply _merging_ this code will not change any framework behavior. It will continue to use the standard library resolver paths, leak DNS, etc. In order to trigger any of the changes this PR implements, the `@@resolver` class variable has to be _manually_ (for now) set by the user. Those who know what this is can turn it on (i do this during framework init), test it, gain its benefits, and catch whatever bugs rmain... everyone else will see no change to their framework/rex-socket experience.

sempervictus:
Merge conflict in here for where i fixed a call to ::Socket and upstream did the same but with `#resolv_nbo`. The upstream fix still works fine with these changes because `@@resolver` in `Rex::Socket` still picks up the request - verified in IRB.

smcintyre-r7:
Tested this a bit in isolation with the recommended pry statements and then I tested a couple of modules with a custom resolver set.

I proposed some minor adjustments in sempervictus#1.

Issue #29, opened 2021-03-14
Server Name Indication for Rex::Socket::Parameters needed?


T35R6braPwgDJKq:
If deduced correctly an SNI parameter is needed: https://github.com/rapid7/metasploit-framework/issues/14653?_pjax=%23js-repo-pjax-container#issuecomment-798914673

smcintyre-r7:
That issue was closed, so I don't think there's any issue here that needs to be resolved. If that's incorrect, just let me know and we can reopen this ticket to discuss details.

robhinds:
Hi @smcintyre-r7 I believe a wider solution to providing SNI (at a lower level than specific module - e.g. Rex::Socket::Tcp etc, as @T35R6braPwgDJKq  mentions in linked ticket) is a good idea? 

I have had similar SNI issues using other modules so assume being able to set it more globally would help - I had the same problem as https://github.com/rapid7/metasploit-framework/issues/14653 but with the `scanner/http/ssl` module, for example. I applied a similar patch as ssl_intercept which resolved it. Also having issues with robot_txt which I believe are also related to not passing through SNI.

(apologies if this is my oversight, I don't normally use Ruby and am new to Metasploit!)

smcintyre-r7:
If I understand correctly, it sounds like the changes implemented in rapid7/metasploit-framework#14911 that closed rapid7/metasploit-framework#14653 should be moved into Rex::Sockets so that all modules creating SSL TCP client connections can properly handle SNI.

robhinds:
Yes, I think so @smcintyre-r7  - I think maybe an advanced option to provide an SNI value that then gets passed down to the socket creation step.

I can see there is code that attempts to set the hostname for SNI here https://github.com/rapid7/rex-socket/blob/master/lib/rex/socket/ssl_tcp.rb#L129 but in my case that is not working, so having an option to explicitly provide that value seems like a good solution.

cyberbutler:
All,

[...]

PR #45, opened 2022-04-25
Allow explicitly setting peer hostname

adfoster-r7:
Closes https://github.com/rapid7/rex-socket/issues/29
Related to https://github.com/rapid7/metasploit-framework/pull/16378
Framework pull request: https://github.com/rapid7/rex-socket/pull/45
rex-socket pull request: https://github.com/rapid7/rex-socket/pull/45

Allow for SNI information to be set explicitly by the caller

### DIFF
--- a/lib/rex/socket/ssl_tcp.rb
+++ b/lib/rex/socket/ssl_tcp.rb
@@ -123,10 +123,14 @@ begin
     # Tie the context to a socket
     self.sslsock = OpenSSL::SSL::SSLSocket.new(self, self.sslctx)
 
-    # If peerhost looks like a hostname, set the undocumented 'hostname'
+    # If peerhostname is set, or if hostname looks like a hostname, set the undocumented 'hostname'
     # attribute on sslsock, which enables the Server Name Indication (SNI)
     # extension
-    self.sslsock.hostname = self.peerhost if !Rex::Socket.dotted_ip?(self.peerhost)
+    if self.peerhostname
+      self.sslsock.hostname = self.peerhostname
+    else !Rex::Socket.dotted_ip?(self.peerhost)
+      self.sslsock.hostname = self.peerhost
+    end
 
     # Force a negotiation timeout
     begin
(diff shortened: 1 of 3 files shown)


lib/rex/socket/ssl_tcp.rb
```diff
@@ -123,10 +123,14 @@ def initsock_with_ssl_version(params, version)
     # Tie the context to a socket
     self.sslsock = OpenSSL::SSL::SSLSocket.new(self, self.sslctx)
 
-    # If peerhost looks like a hostname, set the undocumented 'hostname'
+    # If peerhostname is set, or if hostname looks like a hostname, set the undocumented 'hostname'
```
smcintyre-r7, line 126:
```suggestion
    # If peerhostname is set, or if peerhost looks like a hostname, set the undocumented 'hostname'
```

Happy to ninja commit that doc string change.

smcintyre-r7 (approved):
Confirmed I was able to reproduce the original issue and that this patch fixes it. The code all looks great so I'll get this landed.

I used this little test harness to experiment with the changes:
```
$LOAD_PATH.unshift(File.dirname(__FILE__))
require 'rex/socket'

sock = Rex::Socket::SslTcp.create(
  #'PeerHost' => '18.235.227.114',
  'PeerHost' => 'nvd.nist.gov',
  #'PeerHostname' => 'nvd.nist.gov',
  'PeerPort' => 443
)
sock.write("GET / HTTP/1.0\r\n\r\n")
puts sock.read(2048)
sock.close
```

nvd.nist.gov was called out in one of the issues as a site that requires the SNI configuration in order for the socket to be accepted. Using that harness I was able to test the following three cases:

* `Peerhost` IP address -- Crashes as expected, which is the original issue
* `Peerhost` Hostname -- Works as expected, confirming there is no regression
* `Peerhost` IP address with `PeerHostname` -- Works as expected and reflects Metasploit's use case

Thanks @adfoster-r7 I'll get this landed!

PR #48, opened 2022-05-09
Fix typo in ssl_tcp.rb that would allow for dotted IPs being used for self.sslsock.hostname, which should never happen.

### LINKED ISSUE #47: Typo Leads to Dotted IPs being used for self.sslsock.hostname When They Aren't Meant to Be Used
https://github.com/rapid7/rex-socket/blob/master/lib/rex/socket/ssl_tcp.rb#L131 has a bug in that it should be `elsif` not `else`. Otherwise what happens is that the case that is provided will be ignored and we will always trigger the `else` condition if the above `self.peerhostname` condition isn't true. In reality we only want to execute that code if `!Rex::Socket.dotted_ip?(self.peerhost)` evaluates to true, in which case we then execute `self.sslsock.hostname = self.peerhost`. 

In its current case we will be setting `self.sslsock.hostname = self.peerhost` even in cases where `self.peerhost` is a dotted IP address, which defeats the purpose of the check that was added here.

We can confirm this is the case with a simple test bit of code.

```
3.0.2 :016 > if nil
3.0.2 :017 >   print "A"
3.0.2 :018 > else nil
3.0.2 :019 >   print "B"
3.0.2 :020 > end
B => nil 
3.0.2 :021 > if nil
3.0.2 :022 >   print "A"
3.0.2 :023 > elsif nil
3.0.2 :024 >   print "B"
3.0.2 :025 > end
 => nil
```

As you can see in the `elsif` case, both conditions are properly skipped over. However in the `if...else` case, the second condition is executed even though the condition passed to `else` was false, since `else` doesn't expect an argument and therefore will ignore the argument.

gwillcox-r7:
Fix typo in ssl_tcp.rb that would allow for dotted IPs being used for self.sslsock.hostname, which should never happen.

Fixes #47. 

https://github.com/rapid7/rex-socket/blob/master/lib/rex/socket/ssl_tcp.rb#L131 has a bug in that it should be `elsif` not  `else `. Otherwise what happens is that the case that is provided will be ignored and we will always trigger the else condition if the above self.peerhostname condition isn't true. In reality we only want to execute that code  `if !Rex::Socket.dotted_ip?(self.peerhost)` evaluates to true, in which case we then execute `self.sslsock.hostname = self.peerhost`.

In its current case we will be setting `self.sslsock.hostname = self.peerhost` even in cases where `self.peerhost` is a dotted IP address, which defeats the purpose of the check that was added here.

We can confirm this is the case with a simple test bit of code.

```
3.0.2 :016 > if nil
3.0.2 :017 >   print "A"
3.0.2 :018 > else nil
3.0.2 :019 >   print "B"
3.0.2 :020 > end
B => nil 
3.0.2 :021 > if nil
3.0.2 :022 >   print "A"
3.0.2 :023 > elsif nil
3.0.2 :024 >   print "B"
3.0.2 :025 > end
 => nil
```

As you can see in the elsif case, both conditions are properly skipped over. However in the if...else case, the second condition is executed even though the condition passed to else was false, since else doesn't expect an argument and therefore will ignore the argument.

### DIFF
--- a/lib/rex/socket/ssl_tcp.rb
+++ b/lib/rex/socket/ssl_tcp.rb
@@ -128,7 +128,7 @@ begin
     # extension
     if self.peerhostname
       self.sslsock.hostname = self.peerhostname
-    else !Rex::Socket.dotted_ip?(self.peerhost)
+    elsif !Rex::Socket.dotted_ip?(self.peerhost)
       self.sslsock.hostname = self.peerhost
     end
 


gwillcox-r7:
@adfoster-r7 This was introduced with https://github.com/rapid7/rex-socket/pull/45/files#diff-00d228ddd88d3a644f4b3fbc74e790e7acaecc87b23391b6d565dc6fd7e528afR129-R133 so I imagine https://github.com/rapid7/rex-socket/pull/45 would have the testing directions for this. I'll see if I can't pull them in here.

gwillcox-r7:
Might also be related to https://github.com/rapid7/metasploit-framework/pull/16499, if you want to use that as a testing base.

smcintyre-r7 (approved):

smcintyre-r7:
Well I tested this with the same script I used while testing [#45](https://github.com/rapid7/rex-socket/pull/45#pullrequestreview-960735174). I was able to trace execution with a debugger and confirm the branch was always being taken.

I was *not* able to reproduce a condition however where the branch should not have been taken that taking it cause the connection to fail. The branch intended to prevent setting `#hostname` to an IP address, however doing so didn't appear to break anything. This was however using the default setting that disabled verification, but that didn't prevent SNI exceptions from being raised before.

PR #52, opened 2022-09-13
Fix openssl3 unsafe legacy renegotiation disabled error

adfoster-r7:
Relates to: https://github.com/rapid7/metasploit-framework/issues/16954

This PR fixes the following error in Metasploit-framework when the host machine has OpenSSL 3:

```ruby
msf6 auxiliary(scanner/http/title) > run https://www.courts.ie/

[*] Error: 137.191.225.100: OpenSSL::SSL::SSLError SSL_connect returned=1 errno=0 peeraddr=137.191.225.100:443 state=error: unsafe legacy renegotiation disabled
[*] Scanned 1 of 1 hosts (100% complete)
[*] Auxiliary module execution completed
```

### OpenSSL 1.x.x

Default SSLContext flags - used for Net HTTP
```ruby
2.7.6 :027 > flags = OpenSSL::SSL::SSLContext::DEFAULT_PARAMS[:options]; pp OpenSSL::SSL.constants.select { |constant| constant.to_s.start_with?('OP_') }.map { |c| [c, flags & OpenSSL::SSL.const_get(c) > 0] }.select(&:last); nil
[[:OP_ALL, true],
 [:OP_LEGACY_SERVER_CONNECT, true],
 [:OP_TLSEXT_PADDING, true],
 [:OP_SAFARI_ECDHE_ECDSA_BUG, true],
 [:OP_NO_COMPRESSION, true],
 [:OP_CRYPTOPRO_TLSEXT_BUG, true]]
 => nil 
```

OP_ALL - used internally by rex-socket
```ruby
3.0.2 :007 > flags = OpenSSL::SSL::OP_ALL; pp OpenSSL::SSL.constants.select { |constant| constant.to_s.start_with?('OP_') }.map { |c| [c, flags & OpenSSL::SSL.const_get(c) > 0] }.select(&:last); nil
[[:OP_DONT_INSERT_EMPTY_FRAGMENTS, true],
 [:OP_CRYPTOPRO_TLSEXT_BUG, true],                                                                                                                                                                     
 [:OP_ALL, true],
 [:OP_LEGACY_SERVER_CONNECT, true],
 [:OP_TLSEXT_PADDING, true],
 [:OP_SAFARI_ECDHE_ECDSA_BUG, true]]
```

### OpenSSL 3

Default SSLContext flags - used for Net HTTP - missing OP_LEGACY_SERVER_CONNECT
```ruby
>> flags = OpenSSL::SSL::SSLContext::DEFAULT_PARAMS[:options]; pp OpenSSL::SSL.constants.select { |constant| constant.to_s.start_with?('OP_') }.map { |c| [c, flags & OpenSSL::SSL.const_get(c) > 0] }.select(&:last); nil
[[:OP_ALL, true],
 [:OP_TLSEXT_PADDING, true],                                                                                                                                                                                               
 [:OP_SAFARI_ECDHE_ECDSA_BUG, true],
 [:OP_NO_COMPRESSION, true],
 [:OP_CRYPTOPRO_TLSEXT_BUG, true],
 [:OP_ALLOW_UNSAFE_LEGACY_RENEGOTIATION, true]]
```

OP_ALL - used internally by rex-socket

```ruby
irb(main):003:0> flags = OpenSSL::SSL::OP_ALL; pp OpenSSL::SSL.constants.select { |constant| constant.to_s.start_with?('OP_') }.map { |c| [c, flags & OpenSSL::SSL.const_get(c) > 0] }.select(&:last); nil
[[:OP_TLSEXT_PADDING, true],
 [:OP_SAFARI_ECDHE_ECDSA_BUG, true],                                                                                                                                                                       
 [:OP_DONT_INSERT_EMPTY_FRAGMENTS, true],                                                 
 [:OP_CRYPTOPRO_TLSEXT_BUG, true],                                                        
 [:OP_ALL, true]]                                                                         
=> nil
```

Without this change rex-socket connections break:

It's also possible to set `OP_ALLOW_UNSAFE_LEGACY_RENEGOTIATION` - which is technically what the error suggests to configure

Raw example:
```ruby
require 'socket'
require 'net/http'
require 'uri'

# Flag which was removed from OP_ALL in OpenSSL 3
# OpenSSL::SSL::SSLContext::DEFAULT_PARAMS[:options] |= OpenSSL::SSL::OP_LEGACY_SERVER_CONNECT
# Alternatively - enabling the flag which appears in the error
# OpenSSL::SSL::SSLContext::DEFAULT_PARAMS[:options] |= OpenSSL::SSL::OP_ALLOW_UNSAFE_LEGACY_RENEGOTIATION

puts Net::HTTP.get(URI('https://www.courts.ie/'))
```

Without changes the following error occurs:

```
/var/lib/gems/3.0.0/gems/net-protocol-0.1.3/lib/net/protocol.rb:46:in `connect_nonblock': SSL_connect returned=1 errno=0 peeraddr=137.191.225.100:443 state=error: unsafe legacy renegotiation disabled (OpenSSL::SSL::SSLError)
        from /var/lib/gems/3.0.0/gems/net-protocol-0.1.3/lib/net/protocol.rb:46:in `ssl_socket_connect'
        from /usr/lib/ruby/3.0.0/net/http.rb:1038:in `connect'
        from /usr/lib/ruby/3.0.0/net/http.rb:970:in `do_start'
        from /usr/lib/ruby/3.0.0/net/http.rb:959:in `start'
        from /usr/lib/ruby/3.0.0/net/http.rb:621:in `start'
        from /usr/lib/ruby/3.0.0/net/http.rb:496:in `get_response'
        from /usr/lib/ruby/3.0.0/net/http.rb:467:in `get'
        from foo.rb:6:in `<main>'
```

### DIFF
--- a/lib/rex/socket/ssl_tcp.rb
+++ b/lib/rex/socket/ssl_tcp.rb
@@ -109,6 +109,7 @@ begin
     end
 
     self.sslctx.options = OpenSSL::SSL::OP_ALL
+    self.sslctx.options |= OpenSSL::SSL::OP_LEGACY_SERVER_CONNECT
 
     if params.ssl_cipher
       self.sslctx.ciphers = params.ssl_cipher


lib/rex/socket/ssl_tcp.rb
```diff
@@ -109,6 +109,7 @@ def initsock_with_ssl_version(params, version)
     end
 
     self.sslctx.options = OpenSSL::SSL::OP_ALL
+    self.sslctx.options |= OpenSSL::SSL::OP_LEGACY_SERVER_CONNECT
```
adfoster-r7, line 112: We'll actually need something similar to fix the raw Net:HTTP calls in framework:

```ruby
OpenSSL::SSL::SSLContext::DEFAULT_PARAMS[:options] |= OpenSSL::SSL::OP_LEGACY_SERVER_CONNECT
```

adfoster-r7: Which looks like folk are using Net HTTP for these scenarios https://github.com/rapid7/metasploit-framework/issues/8705

jmartin-r7:
This looks reasonable as long as this constant is available in all the project's supported Ruby versions.

lib/rex/socket/ssl_tcp.rb
adfoster-r7: @jmartin-r7 Any preference on this one? I'm not a fan of globally changing OpenSSL::SSL::SSLContext, as this could impact other things that I'm not aware of. Even if it was the previous behavior. 

It's required for Net HTTP to work as intended:

```ruby
require 'socket'
require 'net/http'
require 'uri'

# Flag which was removed from OP_ALL in OpenSSL 3
# OpenSSL::SSL::SSLContext::DEFAULT_PARAMS[:options] |= OpenSSL::SSL::OP_LEGACY_SERVER_CONNECT
# Alternatively - enabling the flag which appears in the error
# OpenSSL::SSL::SSLContext::DEFAULT_PARAMS[:options] |= OpenSSL::SSL::OP_ALLOW_UNSAFE_LEGACY_RENEGOTIATION

puts Net::HTTP.get(URI('https://www.courts.ie/'))
```

If we introduce it, I could add it to msfenv https://github.com/rapid7/metasploit-framework/blob/b06b39dc71181aa425ccfd07f7b29674b7090350/lib/msfenv.rb#L4-L8 - but that could potentially impact edgecase scenarios in Pro that I'm not aware of

jmartin-r7: From docs [here](https://linux.die.net/man/3/ssl_set_options):
>Patched OpenSSL client and unpatched server.
>
>If the option SSL_OP_LEGACY_SERVER_CONNECT or SSL_OP_ALLOW_UNSAFE_LEGACY_RENEGOTIATION is set then initial connections and renegotiation between patched OpenSSL clients and unpatched servers succeeds. If neither option is set then initial connections to unpatched servers will fail.
The option SSL_OP_LEGACY_SERVER_CONNECT is currently set by default even though it has security implications: otherwise it would be impossible to connect to unpatched servers (i.e. all of them initially) and this is clearly not acceptable. Renegotiation is permitted because this does not add any additional security issues: during an attack clients do not see any renegotiations anyway.
>
>As more servers become patched the option SSL_OP_LEGACY_SERVER_CONNECT will not be set by default in a future version of OpenSSL.
>
>OpenSSL client applications wishing to ensure they can connect to unpatched servers should always set SSL_OP_LEGACY_SERVER_CONNECT
>
>OpenSSL client applications that want to ensure they can not connect to unpatched servers (and thus avoid any security issues) should always clear SSL_OP_LEGACY_SERVER_CONNECT using SSL_CTX_clear_options() or SSL_clear_options().
>
>The difference between the SSL_OP_LEGACY_SERVER_CONNECT and SSL_OP_ALLOW_UNSAFE_LEGACY_RENEGOTIATION options is that SSL_OP_LEGACY_SERVER_CONNECT enables initial connections and secure renegotiation between OpenSSL clients and unpatched servers only, while SSL_OP_ALLOW_UNSAFE_LEGACY_RENEGOTIATION allows initial connections and renegotiation between OpenSSL and unpatched clients or servers.

`SSL_OP_ALLOW_UNSAFE_LEGACY_RENEGOTIATION` might be the better option if we set globally to ensure web servers put up by modules or services are compatible with clients from as many targets as possible however maybe that also should be in the server config vs global.

adfoster-r7:
Looks like it was introduced 5 years ago - https://github.com/ruby/openssl/commit/b44ab7f7e7e1c5f0cf618a347579d090b390103d#diff-09f822c26289f5347111795ca22ed7ed1cfadd6ebd28f987991d1d414eef565aR2674


And is pretty supported :+1:

```
$ docker run -it --rm -w $(pwd) -v $(pwd):$(pwd) ruby:2.5-alpine /bin/sh -c 'ruby -r openssl -e "puts OpenSSL::SSL::OP_LEGACY_SERVER_CONNECT.inspect"'
4
```

jmartin-r7 (approved):

lib/rex/socket/ssl_tcp.rb
adfoster-r7: I'm still on the fence about flipping the `SSL_OP_ALLOW_UNSAFE_LEGACY_RENEGOTIATION` flag on; I can look at that separately - was just hoping to meet feature parity between openssl1 and 3 for now as a first pass :+1:

PR #56, opened 2023-01-18
Stream Control Transmission Protocol (SCTP)

sempervictus:
RFC: https://www.rfc-editor.org/rfc/rfc4960.html
    
Define constants for SCTP in ::Socket namespace for IPPPROTO
and SOL.
Create param definitions and socket marshalling using the newly
defined constants.
Implement stream-server mechanics for SctpServer from TcpServer
extending the clients with Rex::Socket::Sctp upon connection.

Testing:
```ruby
svr = Rex::Socket::SctpServer.create(
  'LocalHost' => '127.0.0.1', 'LocalPort' => 8888
)
csock = nil
Thread.new { csock = svr.accept }
cli = Rex::Socket::Sctp.create(
  'PeerHost' => '127.0.0.1', 'PeerPort' => 8888
)
cli.write("test\n")
csock.gets
$stdout.write(File.read('/proc/net/sctp/eps'))
```
    
Notes:
  No idea how/whether this is going to work on non-Linux hosts
  Meterpreters will need SCTPs implementation to pivot, along with the relevant encapsulation and relay mechanics within TLVs.

### DIFF
--- a/lib/rex/socket/comm/local.rb
+++ b/lib/rex/socket/comm/local.rb
@@ -6,6 +6,8 @@ require 'rex/socket/tcp'
 require 'rex/socket/ssl_tcp'
 require 'rex/socket/ssl_tcp_server'
 require 'rex/socket/udp'
+require 'rex/socket/sctp'
+require 'rex/socket/sctp_server'
 require 'rex/socket/ip'
 require 'timeout'
 
@@ -34,6 +36,8 @@ class Rex::Socket::Comm::Local
         return create_by_type(param, ::Socket::SOCK_STREAM, ::Socket::IPPROTO_TCP)
       when 'udp'
         return create_by_type(param, ::Socket::SOCK_DGRAM, ::Socket::IPPROTO_UDP)
+      when 'sctp'
+        return create_by_type(param, ::Socket::SOCK_STREAM, ::Socket::IPPROTO_SCTP)
       when 'ip'
         return create_ip(param)
       else
@@ -194,15 +198,21 @@ class Rex::Socket::Comm::Local
       sock.setsockopt(::Socket::SOL_SOCKET, ::Socket::SO_BROADCAST, true)
     end
 
-    # If a server TCP instance is being created...
+    # If a server instance is being created...
     if param.server?
       sock.listen(256)
 
       if !param.bare?
-        klass = Rex::Socket::TcpServer
-        if param.ssl
-          klass = Rex::Socket::SslTcpServer
-        end
+        if param.proto == 'tcp'
+          klass = Rex::Socket::TcpServer
+          if param.ssl
+            klass = Rex::Socket::SslTcpServer
+          end
+        elsif param.proto == 'sctp'
+            klass = Rex::Socket::SctpServer
+        else
+          raise Rex::BindFailed.new(param.localhost, param.localport), caller
+        end  
         sock.extend(klass)
 
         sock.initsock(param)
@@ -308,6 +318,9 @@ class Rex::Socket::Comm::Local
             sock.initsock(param)
           when 'udp'
             sock.extend(Rex::Socket::Udp)
+            sock.initsock(param)
+          when 'sctp'
+            sock.extend(Rex::Socket::Sctp)
             sock.initsock(param)
         end
       end
--- a/lib/rex/socket/sctp.rb
+++ b/lib/rex/socket/sctp.rb
@@ -1,0 +1,66 @@
+# -*- coding: binary -*-
+require 'rex/socket'
+require 'rex/io/stream'
+
+class ::Socket
+  IPPROTO_SCTP = 132
+  SOL_SCTP     = 132
+end
+
+###
+#
+# This class provides methods for interacting with a SCTP client connection.
+#
+###
+module Rex::Socket::Sctp
+
+  include Rex::Socket
+  include Rex::IO::Stream
+
+  ##
+  #
+  # Factory
+  #
+  ##
+
+  #
+  # Creates the client using the supplied hash.
+  #
+  # @see create_param
+  # @see Rex::Socket::Parameters.from_hash
+  def self.create(hash = {})
+    hash['Proto'] = 'sctp'
+    self.create_param(Rex::Socket::Parameters.from_hash(hash))
+  end
+
+  #
+  # Wrapper around the base socket class' creation method that automatically
+  # sets the parameter's protocol to SCTP.
+  #
+  def self.create_param(param)
+    param.proto = 'sctp'
+    Rex::Socket.create_param(param)
+  end
+
+  ##
+  #
+  # Stream mixin implementations
+  #
+  ##
+
+  #
+  # Calls shutdown on the SCTP connection.
+  #
+  def shutdown(how = ::Socket::SHUT_RDWR)
+    begin
+      return (super(how) == 0)
+    rescue ::Exception
+    end
+  end
+
+  # returns socket type
+  def type?
+    return 'sctp'
+  end
+
+end
--- a/lib/rex/socket/sctp_server.rb
+++ b/lib/rex/socket/sctp_server.rb
@@ -1,0 +1,75 @@
+# -*- coding: binary -*-
+require 'rex/socket'
+require 'rex/socket/sctp'
+require 'rex/io/stream_server'
+
+###
+#
+# This class provides methods for interacting with a SCTP server.  It
+# implements the Rex::IO::StreamServer interface.
+#
+###
+module  Rex::Socket::SctpServer
+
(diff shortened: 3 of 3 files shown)


adfoster-r7:
Looks like you're pulling in some commits that are already landed to master; I think a rebase will fix that

sempervictus:
Ping @hdm and skape (can't seem to find a GH handle which is mildly ironic): i did this in an hour this morning and it dawns on me that there had to be some reason why the founders _didn't_ - obviously there aren't any comments to find on code that doesn't exist, so hoping you might recall whether this was discussed and why it wasn't implemented if it was.
It works in naive testing, seems to use the correct L4 stack, and operates like TCP so it probably worked before i rewired the Stream/IO abstractions for stateless sockets. Is there some elephant in the room which i'm not seeing (other than Windows not having SCTP sockets natively)?

Ping @zeroSteiner, @OJ, @timwr regarding relevant meterpreter types and discussion re implementing the pivot infrastructure in the various meterps you gents maintain.

Far as SCTP compatibility for Framework-side, looks like SCTPLabs has a usermode SCTP implementation which is cross-platform @ https://github.com/sctplab/usrsctp

sempervictus:
> Looks like you're pulling in some commits that are already landed to master; I think a rebase will fix that

Pardon, all set.

sempervictus:
@adfoster-r7 - we gotta land this for the framework side (new shells!!) to get reviewed. Who's taking this one? You? Come at me bro! :grin:

adfoster-r7:
Looks like SCTP wouldn't work on windows:

```
irb(main):003:0> s = Socket.new(Socket::AF_INET, Socket::SOCK_STREAM, 132)
(irb):3:in `initialize': The requested protocol has not been configured into the system, or no implementation for it exists. - socket(2) (Errno::EPROTONOSUPPORT)
        from (irb):3:in `new'
        from (irb):3:in `<main>'
        from C:/Ruby30-x64/lib/ruby/gems/3.0.0/gems/irb-1.3.6/exe/irb:11:in `<top (required)>'
        from C:/Ruby30-x64/bin/irb:25:in `load'
        from C:/Ruby30-x64/bin/irb:25:in `<main>'
```

Or Mac:

```
2.7.2 :013 > require 'socket'
 => true 
2.7.2 :014 > s = Socket.new(Socket::AF_INET, Socket::SOCK_STREAM, 132)
Traceback (most recent call last):
        23: from /Users/user/.rvm/gems/ruby-2.7.2/bin/ruby_executable_hooks:22:in `<main>'           
        22: from /Users/user/.rvm/gems/ruby-2.7.2/bin/ruby_executable_hooks:22:in `eval'             
        21: from /Users/user/.rvm/gems/ruby-2.7.2/bin/irb:23:in `<main>'                             
        20: from /Users/user/.rvm/gems/ruby-2.7.2/bin/irb:23:in `load'                               
        19: from /Users/user/.rvm/gems/ruby-2.7.2/gems/irb-1.6.2/exe/irb:11:in `<top (required)>'    
         2: from (irb):14:in `<main>'                                                                    
         1: from (irb):14:in `new'                                                                       
(irb):14:in `initialize': Protocol not supported - socket(2) (Errno::EPROTONOSUPPORT)
```

But I was able to create a socket for the linux-y boxes I checked, such as ubuntu 2016, and centos 7, alpine 3.14

```c
#include <stdio.h>
#include <sys/socket.h>

int main() {
  int listenfd;

  listenfd = socket(AF_INET, SOCK_STREAM, 132);

  printf("fd: %d\n", listenfd);
}
```

It would be great if there was wider support; but it looks like it's still a lot of targets that it could work against

lib/rex/socket/sctp.rb
```diff
@@ -0,0 +1,66 @@
+# -*- coding: binary -*-
+require 'rex/socket'
+require 'rex/io/stream'
+
+class ::Socket
+  IPPROTO_SCTP = 132
+  SOL_SCTP     = 132
+end
```
adfoster-r7, line 8: Thoughts on adding these constants to `::Rex::Socket` instead?

lib/rex/socket/sctp.rb
```diff
@@ -0,0 +1,66 @@
...
+  #
+  def shutdown(how = ::Socket::SHUT_RDWR)
+    begin
+      return (super(how) == 0)
+    rescue ::Exception
+    end
+  end
+
+  # returns socket type
+  def type?
```
adfoster-r7, line 62: I would've expected this to return a boolean; but looks like it's an existing convention to return the socket type string

lib/rex/socket/comm/local.rb
```diff
@@ -194,15 +198,21 @@ def self.create_by_type(param, type, proto = 0)
...
-        end
+        if param.proto == 'tcp'
+          klass = Rex::Socket::TcpServer
+          if param.ssl
+            klass = Rex::Socket::SslTcpServer
+          end
+        elsif param.proto == 'sctp'
+            klass = Rex::Socket::SctpServer
+        else
+          raise Rex::BindFailed.new(param.localhost, param.localport), caller
```
adfoster-r7, line 214: not a blocker: I think this is the right exception to raise here, but I worry it might be hard to debug if this code path ever triggers as it's probably caught and silently ignored in some cases in framework

adfoster-r7 (commented):
I think we if we move the constants to rex socket, I'd be good with these changes - it's mostly copied from the existing tcp files

lib/rex/socket/sctp.rb
sempervictus: At the end of the day, `Rex::Socket`s are still `Socket`s so i figure the constant belongs with the rest of them. No objection to moving if we have a good reason.

lib/rex/socket/comm/local.rb
sempervictus: If that is the case, (IMO) we should fix it higher up in the stack where it is erroneously ignored.
I look at `Rex` as the "kernel" of Metasploit which needs to be consistent and informative as much as possible, especially going down to the core (Socket, IO, Event, locking, all that fun junk). That said, `Error`s are `Object`s too and we can do whatever we want with them - like mixing in/extending with an ancestor which will increase chances of trapping on `Socket` errors...

lib/rex/socket/sctp.rb
adfoster-r7: The rationale behind the comment was - we shouldn't patch core files we don't own, that's why `::Rex::Socket` exists in the first place right?

There's also load order issues if a module ends up depending on this patched `::Socket` constant, but this file hasn't been loaded etc

sempervictus:
Kind of? `Rex::Socket` is really a factory for the dynamic dispatch of API-consistent "virtual wiring" but it _uses_ `::Socket` for "everything" under the skin. If you're worried about namespace corruption, just take a look at `.class.ancestors` in almost any MSF object - some of the gems we pull in do things that make my work look downright pristine :wink: 
IMO, I'm not so much corrupting someone else's namespace as i'm fixing a missing element in their implementation - that's a real Socket class per the RFC, and they just neglected to define the constant.

lib/rex/socket/sctp.rb
sempervictus: Kind of? Rex::Socket is really a factory for the dynamic dispatch of API-consistent "virtual wiring" but it uses ::Socket for "everything" under the skin. If you're worried about namespace corruption, just take a look at .class.ancestors in almost any MSF object - some of the gems we pull in do things that make my work look downright pristine wink
IMO, I'm not so much corrupting someone else's namespace as i'm fixing a missing element in their implementation - that's a real Socket class per the RFC, and they just neglected to define the constant.

sempervictus:
Regarding OS support, from what i can tell, it's any moderately modern Linux (Android is an open question), any FreeBSD after 7 (which to me implies HBSD as well, but i'll test that theory to make sure), as well as Solaris 10+.
Macs _had_ an NKE up to 10.13, but i think that the way to go nowadays would be https://github.com/sctplab/usrsctp. Same goes for Windows, and would need to be implemented inside Meterpreter/Mettle since its a binary object (unless someone knows of a way to carry those into Py/PHP/Java execution space correctly).

lib/rex/socket/sctp.rb
adfoster-r7: I'll treat this as not a blocker, but I'm not a massive fan of the pattern

sempervictus:
Thanks boss.
Given that we're one of the biggest Ruby projects out there, maybe its not a bad idea for us to reach out to whatever Ruby's version of a steering committee is and ask them what the _right_ way to do things is from their point of view. Commits going in here are not likely to be touched (or seen) by another soul for quite a while, but being infrastructure code, everything runs atop it...

PR #61, opened 2023-06-02
Add in function to convert IPv6 addresses to IPv6 URL compatible format

gwillcox-r7:
This uses https://en.wikiversity.org/wiki/IPv6#Use_in_URLs and the information in several RFCs which specify that IPv6 addresses in URLs should be enclosed in `[ ]` and makes a function in Rex-Socket that makes it easy to identify if an address is an IPv6 address and convert it into a URL compatible string.

This came up after some discussion with @smcintyre-r7 about the lack of ability to do this in Metasploit and the increased need to support IPv6 in Metasploit.

### DIFF
--- a/lib/rex/socket.rb
+++ b/lib/rex/socket.rb
@@ -157,7 +157,33 @@ module Socket
     (support_ipv6? && addr =~ MATCH_IPV6) || (addr =~ MATCH_IPV4)
   end
 
+  # Checks to see if an address is an IPv6 address and if so, converts it into its
+  # square bracket format for addressing as noted in RFC 6874 which states that an IPv6
+  # address literal in a URL is always embedded between [ and ]. Please also refer to
+  # RFC5952, RFC3986, and RFC6874 for more info.
   #
+  # RFC3986 section 3.2.2 specifically notes "A host identified by an Internet Protocol literal address, version 6
+  # [RFC3513] or later, is distinguished by enclosing the IP literal
+  # within square brackets ("[" and "]").  This is the only place where
+  # square bracket characters are allowed in the URI syntax."
+  #
+  # RFC6874 reinforces this in section 2 where it notes "In a URI, a literal IPv6 address
+  # is always embedded between '[' and ']'".
+  #
+  # @param host [String] IP address or hostname to convert to a URI authority.
+  # @param port [Integer] Port number to include within the URI authority.
+  # @return [String] Returns the URI authority string.
+  # @raise [ArgumentError] This function will raise an ArgumentError if the host parameter is not a String.
+  def self.to_authority(host, port=nil)
+    unless host.kind_of?(String)
+      raise ArgumentError.new("Expected a string for the host parameter!")
+    end
+    host = "[#{host}]" if is_ipv6?(host)
+    host += ":#{port}" if port
+    host
+  end
+
+
   # Return true if +addr+ is within the ranges specified in RFC1918, or
   # RFC5735/RFC3927
   #


lib/rex/socket.rb
```diff
@@ -157,7 +157,27 @@ def self.dotted_ip?(addr)
...
+  # @param addr [String] IPv6 address to convert to URL format.
+  # @return [String, nil] Returns the converted address on success, or nil on failure.
+  def self.convert_ipv6_to_url_fmt(addr)
+    if is_ipv6?(addr)
+      "[#{addr}]"
+    else
+      nil
+    end
+  end
+
```
smcintyre-r7, line 179: This makes a few changes:

1. Renames it to `.to_authority` (see [RFC3986 Section 3](https://datatracker.ietf.org/doc/html/rfc3986#section-3)). Alternatively you could call it `.to_uri_authority` I guess to keep it inline with the RFC3986 verbiage and be more explicit but then that's 4 more character to type and the idea here is to make this more convenient.
2. Takes the port because it's only necessary to enclose the address with brackets when a port is included. If the port is `Nil` the host is returned as is, again for the caller's convenience.
3. Always returns a string because this will make it easier to use. The caller won't have to check if the conversion took place, they can simply use the result.

```suggestion
  # Checks to see if an address is an IPv6 address and if so, converts it into its
  # square bracket format for addressing as noted in RFC 6874 which states that an IPv6
  # address literal in a URL is always embedded between [ and ]. Please also refer to
  # RFC5952 and RFC3986 for more info.
  #
  # RFC3986 section 3.2.2 specifically notes "A host identified by an Internet Protocol literal address, version 6
  # [RFC3513] or later, is distinguished by enclosing the IP literal
  # within square brackets ("[" and "]").  This is the only place where
  # square bracket characters are allowed in the URI syntax."
  #
  # @param host [String] IP address or hostname to convert to a URI authority.
  # @param port [Integer] Port number to include within the URI authority.
  # @return [String] Returns the URI authority string.
  def self.to_authority(host, port=nil)
    host = "[#{host}]" if port && is_ipv6?(host)
    host += ":#{port}" if port
    host
  end
```

smcintyre-r7 (changes requested):

lib/rex/socket.rb
gwillcox-r7: Suggested change implemented in https://github.com/rapid7/rex-socket/pull/61/commits/07ebf46746ac39e0f1079c9af2f71ee9a608deea.

lib/rex/socket.rb
```diff
@@ -160,19 +160,23 @@ def self.dotted_ip?(addr)
...
+  # RFC6874 reinforces this in section 2 where it notes "In a URI, a literal IPv6 address
+  # is always embedded between '[' and ']'".
+  #
   # @param host [String] IP address or hostname to convert to a URI authority.
   # @param port [Integer] Port number to include within the URI authority.
   # @return [String] Returns the URI authority string.
   def self.to_authority(host, port=nil)
-    host = "[#{host}]" if port && is_ipv6?(host)
-    host += ":#{port}" if port && host.kind_of?(String)
+    return host unless host.kind_of?(String)
```
smcintyre-r7, line 177: This conflicts with the docs that states a string is always returned and that `host` should be a string. If you want to account for potential issues where `host` is not a string, then you should raise an [ArgumentError](https://ruby-doc.org/core-2.5.0/ArgumentError.html) exception stating that `host` must be a string.

gwillcox-r7: Good point, fixed this with https://github.com/rapid7/rex-socket/pull/61/commits/6e67364664cfbba3d0ffb67947a5d49bb43598e3 so that we throw the ArgumentError on an invalid `host` input as you recommended.

smcintyre-r7 (approved):
Changes LGTM, thanks!

```
[1] pry(main)> require 'rex/socket'
=> true
[2] pry(main)> Rex::Socket.to_authority('::1', 80)
=> "[::1]:80"
[3] pry(main)> Rex::Socket.to_authority('::1')
=> "[::1]"
[4] pry(main)> Rex::Socket.to_authority('localhost')
=> "localhost"
[5] pry(main)> Rex::Socket.to_authority('localhost', 80)
=> "localhost:80"
[6] pry(main)> Rex::Socket.to_authority('localhost', 0)
=> "localhost:0"
[7] pry(main)> Rex::Socket.to_authority('localhost', nil)
=> "localhost"
[8] pry(main)> Rex::Socket.to_authority('127.0.0.1', nil)
=> "127.0.0.1"
[9] pry(main)> Rex::Socket.to_authority('127.0.0.1', 0)
=> "127.0.0.1:0"
[10] pry(main)> Rex::Socket.to_authority('127.0.0.1', 800)
=> "127.0.0.1:800"
[11] pry(main)> Rex::Socket.to_authority(1, 800)
Rex::ArgumentError: An invalid argument was specified. Expected a string for the host parameter!
from /home/smcintyre/Repositories/rex-socket.pr/lib/rex/socket.rb:179:in `to_authority'
[12] pry(main)> 
```

PR #65, opened 2024-02-09
Add .is_name? And Strengthen .is_ip_addr? Validation

zeroSteiner:
This makes two changes. The first is it adds a `.is_name?` regex that can be used to test if a string *appears* to be a valid DNS hostname. It does not perform any DNS resolution, it only checks that it appears to be well constructed, e.g. it does not contain any bad characters like `this;&isNotAValid_(Name)`.

This second change is to strengthen the existing `.is_ipv4?` and `.is_ipv6?` tests to ensure that when the result is true, that the same value can be passed to `IPAddr.new` without worrying about an exception. More specifically, due to how whitespace was being handled, strings which contained new lines could evaluate to true, despite not being proper IP addresses. New unit tests were added for this.

Example:
```
irb(main):001:0> $LOAD_PATH.unshift 'lib'; require 'ipaddr'; require 'rex/socket'
=> true
irb(main):002:0> fake_ip_address = "127.0.0.1\nTest"
=> "127.0.0.1\nTest"
irb(main):003:0> IPAddr.new(fake_ip_address) if Rex::Socket.is_ip_addr?(fake_ip_address)
/usr/share/ruby/ipaddr.rb:684:in `in6_addr': invalid address:  (IPAddr::InvalidAddressError)

      raise InvalidAddressError, "invalid address: #{@addr}"
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
	from /usr/share/ruby/ipaddr.rb:620:in `initialize'
	from (irb):3:in `new'
	from (irb):3:in `<main>'
	from /usr/share/gems/gems/irb-1.6.2/exe/irb:11:in `<top (required)>'
	from /usr/bin/irb:25:in `load'
	from /usr/bin/irb:25:in `<main>'
irb(main):004:0> 
```


sempervictus (approved):
Thanks, looks sane to me.

PR #66, opened 2024-03-25
Fix/is name underscores

zeroSteiner:
Per RFC 2181, DNS names can contain underscores. Since the `.is_name?` class method is intended to inspect the value and tell if it's value for resolution, it should allow underscores in the labels.

Included a test to show that the legitimate case of `_ldap._tcp.msflab.local` now works as intended.


spec/rex/socket_spec.rb
```diff
@@ -342,6 +342,13 @@
        end
      end
 
+    context 'with a name containing underscores' do
```
adfoster-r7, line 345: Not a blocker; would it make sense to rename `is_name?` is `is_dns_name?` or something similar? I intentionally didn't call it on the initial PR as I thought the distinctiong wasn't super importance, but now I wish I had as hostname vs DNS name have very different rules apparently 😄

lib/rex/socket.rb
```diff
@@ -85,7 +85,7 @@ def self.create_ip(opts = {})
   #
 
   # see: https://debugpointer.com/regex/regex-for-dns-name
-  MATCH_DNS_NAME = /^(((?!-))(xn--)?([a-z0-9][a-z0-9\-]{0,59})?[a-z0-9]\.)*(xn--)?([a-z0-9\-]{1,61}|[a-z0-9-]{1,30}\.[a-z]{2,})$/i
+  MATCH_DNS_NAME = /^(((?!-))(xn--)?([a-z0-9_][a-z0-9_\-]{0,59})?[a-z0-9_]\.)*(xn--)?([a-z0-9_\-]{1,61}|[a-z0-9_-]{1,30}\.[a-z]{2,})$/i
```
dwelch-r7, line 88: Should we also be allowing for DNS entries ending with a `.` to be valid? e.g. `google.com.` I'm like 90% sure it's legit to have that work 
seems like an issue ran into by someone already (I'm less convinced by a `.` at the end of a hostname but i think it's still allowed)
https://github.com/rapid7/metasploit-framework/discussions/19001

zeroSteiner: Yeah, I don't know about that. A leading or trailing `.` would imply there's an empty label, which I don't think is valid. When I use wireshark to inspect the query issued by `dig google.com` and `dig google.com.` I can see that `dig` just removes the trailing `.` and normalizes the name to `google.com`. Since this is the socket library and it's pretty closely tied to the lower level networking, I don't think it makes sense to include that type of application-level logic for normalizing the string.

dwelch-r7: I don't think a trailing `.` implies an empty label, as far as I can tell the trailing `.` indicates a FQDN and if not present it's implied to be there?
http://www.dns-sd.org/trailingdotsindomainnames.html

zeroSteiner: Okay, I guess it makes sense because the existing implementation accepts hostnames which also imply that there's resolver behavior in place.

zeroSteiner: Yeah after reading through that, it makes sense to allow a trailing `.`. `.is_name?` already allows hostnames which implies some degree of resolver logic is already in place so it seems reasonable to expand on that. I'll get that added, thanks @dwelch-r7 .

PR #67, opened 2024-11-19
Fix .rex_getaddrinfo inconsistencies

zeroSteiner:
The `.rex_getaddrinfo` method behaves inconsistently with `Addrinfo.getaddrinfo`, which makes `.getaddresses`  inconsistent depending on if a resolver is present or not. When no resolver is present, a numeric address (either decimal or hex, not binary or octal) is specified, it's converted to an IPv4 address:

```
[2] pry(#<Msf::Modules::Exploit__Multi__Ssh__Sshexec::MetasploitModule>)> ::Addrinfo.getaddrinfo('0', 0, ::Socket::AF_UNSPEC, ::Socket::SOCK_STREAM)
=> [#<Addrinfo: 0.0.0.0 TCP (0)>]
```

But when a resolver is present, `.getaddresses` will call `.rex_getaddrinfo` and it falls with a resolution error because the address is not a valid hostname:

```
[10] pry(#<Msf::Modules::Exploit__Multi__Ssh__Sshexec::MetasploitModule>)> Rex::Socket.rex_getaddrinfo('0')
SocketError: getaddrinfo: Name or service not known
from /home/smcintyre/.rvm/gems/ruby-3.1.5@metasploit-framework/gems/rex-socket-0.1.57/lib/rex/socket.rb:996:in `rex_resolve_hostname'
[11] pry(#<Msf::Modules::Exploit__Multi__Ssh__Sshexec::MetasploitModule>)>
```

This came up as an issue that prevents Metasploit's route command from working when invoked as `route add 0 0 -1` to add a default route for all IPv4 addresses through the most recent session when the DNS feature is enabled. This also removes references to `::Net::DNS` in favor of using the constants from `Dnsruby::Types` instead, which was already referenced but not defined as a requirement. Adding Dnsruby as a requirement was needed to make the tests work.

### DIFF
--- a/lib/rex/socket.rb
+++ b/lib/rex/socket.rb
@@ -940,42 +940,30 @@ protected
   # @param resolver [Rex::Proto::DNS::CachedResolver] Resolver to query for the name
   # @return [Array] Array mimicking the native getaddrinfo return type
   def self.rex_getaddrinfo(name, resolver: @@resolver)
-    getaddrinfo = []
+    v4_sockaddrs = []
+    v6_sockaddrs = []
 
     if name =~ /\A\d+\Z/ && name.to_i.between?(0, 0xffffffff)
-      getaddrinfo << Addrinfo.new(
-        self.to_sockaddr(name.to_i, 0),
-        ::Socket::AF_INET,
-        ::Socket::SOCK_STREAM,
-        ::Socket::IPPROTO_TCP
-      )
+      v4_sockaddrs << self.to_sockaddr(name.to_i, 0)
     elsif name =~ /\A0x[0-9a-fA-F]+\Z/ && name.to_i(16).between?(0, 0xffffffff)
-      getaddrinfo << Addrinfo.new(
-        self.to_sockaddr(name.to_i(16), 0),
-        ::Socket::AF_INET,
-        ::Socket::SOCK_STREAM,
-        ::Socket::IPPROTO_TCP
-      )
+      v4_sockaddrs << self.to_sockaddr(name.to_i(16), 0)
+    elsif self.is_ipv4?(name)
+      v4_sockaddrs << self.to_sockaddr(name, 0)
+    elsif self.is_ipv6?(name)
+      v6_sockaddrs << self.to_sockaddr(name, 0)
     else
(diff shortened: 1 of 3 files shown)


adfoster-r7 (approved):

lib/rex/socket.rb
```diff
@@ -939,26 +940,42 @@ def self.rex_gethostbyname(name, resolver: @@resolver)
   # @param resolver [Rex::Proto::DNS::CachedResolver] Resolver to query for the name
   # @return [Array] Array mimicking the native getaddrinfo return type
   def self.rex_getaddrinfo(name, resolver: @@resolver)
-    v4, v6 = self.rex_resolve_hostname(name, resolver: resolver)
-    # Build response array
     getaddrinfo = []
-    v4.each do |a4|
+
+    if name =~ /^\d+$/ && name.to_i.between?(0, 0xffffffff)
```
adfoster-r7, line 945: Not a blocker; `\A` and `\z` are usually used for anchors in Ruby https://stackoverflow.com/questions/3632024/why-do-rubys-regular-expressions-use-a-and-z-instead-of-and

zeroSteiner:
I just noticed that `Addrinfo.getaddrinfo` also handles normal IP addresses. I should really add that here as well. Let me get to that today to fix that inconsistency too.

```
[3] pry(#<Msf::Framework>)> hostname='127.0.0.1'
=> "127.0.0.1"
[4] pry(#<Msf::Framework>)> ::Addrinfo.getaddrinfo(hostname, 0, ::Socket::AF_UNSPEC, ::Socket::SOCK_STREAM)
=> [#<Addrinfo: 127.0.0.1 TCP>]
[5] pry(#<Msf::Framework>)> ::Addrinfo.getaddrinfo('::1', 0, ::Socket::AF_UNSPEC, ::Socket::SOCK_STREAM)
=> [#<Addrinfo: ::1 TCP>]
[6] pry(#<Msf::Framework>)> 
```

zeroSteiner:
Alright this should be ready to go now. I added normal IP address support in the last commit along with tests for it.

adfoster-r7 (approved):

PR #74, opened 2025-04-10
Adds enhanced support for network capture decryption

cgranleese-r7:
This pull request adds enhanced support for network capture decryption. By writing to the sslkeylogfile it enable network capture decryption which is useful to decrypt TLS traffic in wireshark. 

Needs tested with Metasploit-Framework, to do this add the following to the Metasploit-Framework Gemfile:
```
gem "rex-socket", path: "../rex-socket"
```

Then `bundle install`.

## Verification
- [ ] Start msfconsole
- [ ] Use `scanner/http/title`
- [ ] Run `run rhosts=https://www.google.com verbose=true httptrace=true sslkeylogfile=./sslkeylogfile.txt`
- [ ] The module should complete
- [ ] Run `ls -la` and you should now see a file called `sslkeylogfile.txt`

### DIFF
--- a/lib/rex/socket/parameters.rb
+++ b/lib/rex/socket/parameters.rb
@@ -51,6 +51,8 @@ class Rex::Socket::Parameters
   #
   # @option hash [String] 'PeerHost' The remote host to connect to
   # @option hash [String] 'PeerHostname' The unresolved remote hostname, used to specify Server Name Indication (SNI)
+  # @option hash [String] 'SSLKeyLogFile' The SSL key log file path, used for network capture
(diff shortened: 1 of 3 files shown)


lib/rex/socket/parameters.rb
```diff
@@ -302,6 +308,10 @@ def v6?
   # @return [String]
   attr_accessor :peerhostname
 
+  # The SSL key log file path, equivalent to the sslkeylogfile parameter hash
+  # key.
+  attr_accessor :sslkeylogfile
```
cgranleese-r7, line 313:
```suggestion
  # @return [String]
  attr_accessor :sslkeylogfile
```

PR #76, opened 2025-05-07
Add SOCKS5H Support

zeroSteiner:
This adds SOCKS5H support to Rex::Socket, allowing TCP connections established with the local comm to be made through SOCKS5 proxies. The difference between SOCKS5 and SOCKS5H is where hostnames are resolved to DNS. There is a common convention where SOCKS5 proxy clients (including Metasploit) resolve hostnames themselves and issue the IP address to the SOCKS server when connecting. SOCKS5H on the other hand is an unofficial standard whereby the hostname is passed to the SOCKS server, allowing it to be resolved by the proxy instead of Metasploit. 

These changes lay the ground work for Metasploit to support both conventions. There will be a PR to rapid7/metasploit-framework shortly that will take advantage of these changes.

This PR also adds a number of unit tests to show how proxies are handled and ensure resolution does and does not happen when appropriate.

## Proxy Strings
Proxies are now parsed into URI instances instead of arrays. This means Ruby's builtin `URI` class and it's parsing capabilities can be leveraged instead of using a custom solution. In the future this should also allow additional options to be specified in the URI, so if an HTTP proxy requires authentication, it can be specified in the URI. The `//` part of the URi is optional to ensure backwards compatibility, e.g. `http:localhost:8080` will still work but it's normalized to `http://localhost:8080` under the hood so it can be processed as a URI.

## Test Script
This test script will target an SSH server on TCP port 22 which works well as a test case because SSH will send it's banner, showing that the connection was established. This script allows a proxy to be set so a user can interactively observe how their options affect the socket creation. Use wireshark to check the connect request made to the SOCKS server to see if a hostname or IP address was sent.

```ruby
$LOAD_PATH.unshift 'lib'
require 'optparse'
require 'rex/socket'

options = {}
parser = OptionParser.new do |opts|
  opts.banner = "Usage: script.rb TARGET [--proxy PROXY]"

  opts.on("--proxy PROXY", String, "Proxy (optional)") do |proxy|
    options[:proxy] = proxy
  end
end

begin
  parser.parse!
  raise OptionParser::MissingArgument, "TARGET is required" if ARGV.empty?
  options[:target] = ARGV.shift
rescue OptionParser::InvalidOption, OptionParser::MissingArgument => e
  puts e.message
  puts parser
  exit 1
end

puts "Target: #{options[:target]}"
puts "Proxy:  #{options[:proxy]}" if options[:proxy]

client = Rex::Socket::Tcp.create({
  'PeerHost' => options[:target],
  'PeerPort' => 22,
  'SSL'      => false,
  'Proxies'  => options[:proxy]
})
data = client.get_once

$stderr.puts data
client.close
```

### DIFF
--- a/lib/rex/socket/comm/local.rb
+++ b/lib/rex/socket/comm/local.rb
@@ -262,14 +262,15 @@ class Rex::Socket::Comm::Local
         end
 
         ip6_scope_idx = 0
-        ip   = Rex::Socket.getaddress(param.peerhost)
-        port = param.peerport
 
         if param.proxies
           chain = param.proxies.dup
           chain.push(['host',param.peerhost,param.peerport])
-          ip = chain[0][1]
+          ip   = chain[0][1]
           port = chain[0][2].to_i
+        else
+          ip   = Rex::Socket.getaddress(param.peerhost)
+          port = param.peerport
         end
 
         begin
@@ -326,12 +327,13 @@ class Rex::Socket::Comm::Local
         end
       end
 
+      # fixme: handle the chain object here
       if chain.size > 1
         chain.each_with_index {
           |proxy, i|
           next_hop = chain[i + 1]
           if next_hop
-            proxy(sock, proxy[0], next_hop[1], next_hop[2])
+            proxy(sock, proxy[0], next_hop[1], next_hop[2].to_i)
           end
         }
       end
@@ -342,8 +344,6 @@ class Rex::Socket::Comm::Local
         sock.extend(klass)
         sock.initsock(param)
       end
-
-
     end
 
     # Notify handlers that a socket has been created.
--- a/lib/rex/socket/comm/local.rb
+++ b/lib/rex/socket/comm/local.rb
@@ -432,25 +432,21 @@ class Rex::Socket::Comm::Local
         raise Rex::ConnectionProxyError.new(host, port, type, "The proxy returned a non-OK response"), caller
       end
     when Rex::Socket::Proxies::ProxyType::SOCKS4
-      supports_ipv6 = false
-      setup = [4,1,port.to_i].pack('CCn') + Rex::Socket.resolv_nbo(host, supports_ipv6) + Rex::Text.rand_text_alpha(rand(8)+1) + "\x00"
-      size = sock.put(setup)
-      if size != setup.length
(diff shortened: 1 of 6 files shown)


lib/rex/socket/parameters.rb
```diff
@@ -354,7 +355,9 @@ def comm
     best_comm = nil
     # If no comm was explicitly specified, try to use the comm that is best fit
     # to handle the provided host based on the current routing table.
-    if server and localhost
+    if proxies?
+      best_comm = Rex::Socket::Comm::Local
+    elsif server && localhost
```
zeroSteiner, line 360: Right now, only the local comm supports proxied connections. If a comm is specified, the proxy options are silently ignored. This seems like a problem. Solutions are probably to:

1. Warn that proxies will be ignored.
2. Raise an exception that the selected comm doesn't support proxies.
3. Refactor proxy connection login into a mixin that can be included into *any* comm. The comm would be used to establish the connection to the first proxy in the chain, then it'd continue on from there. This would be the most difficult option and is thus left out of scope for this change set.

Kind of related is the fact that proxies are ignored for non-TCP socket types. Also seems like a problem, but supporting proxies in comms won't change that. SOCKS4 and HTTP simply don't support UDP and they never will.

lib/rex/socket/comm/local.rb
```diff
@@ -440,76 +424,49 @@ def self.proxy(sock, type, host, port)
         raise Rex::ConnectionProxyError.new(host, port, type, "Failed to receive a response from the proxy"), caller
       end
 
-      resp = Rex::Proto::Http::Response.new
-      resp.update_cmd_parts(ret.split(/\r?\n/)[0])
+      if (match = ret.match(/HTTP\/.+?\s+(\d+)\s?.+?\r?\n?$/))
+        status_code  = match[1].to_i
+      end
```
zeroSteiner, line 429: This fixed a bug where an exception would be raised if Rex::Socket was used out side the context of Metasploit because `Rex::Proto::Http::Response` is unavailable. The other proxies just do the bare minimum parsing necessary to ensure the connection was established, so that patter was copied here as well. Removing this dependency also means the HTTP proxy can be tested in the specs.

PR #83, opened 2026-05-15
Fix #recvfrom to match stdlib

zeroSteiner:
This is related to #82 but unlike #82, this is a breaking change. It updates the `#recvfrom` signature to match that of the standard library. This was raised first here: https://github.com/rapid7/metasploit-framework/pull/20689#issuecomment-3534578380 and was highlighted by the condition required here: https://github.com/Z6543/ruby_smb/blob/679da4dba6bb4256ff1b79527d5d0c6f8cc6fd76/lib/ruby_smb/nbss/node_status.rb#L215-L224

### DIFF
--- a/lib/rex/socket/udp.rb
+++ b/lib/rex/socket/udp.rb
@@ -135,26 +135,54 @@ module Rex::Socket::Udp
   end
 
   #
-  # Receives a datagram and returns the data and sender address information
-  # as [ data, [address_family, port, host, host] ], matching the format of
-  # stdlib UDPSocket#recvfrom (host appears in both the hostname and numeric
-  # address positions; no reverse-DNS lookup is performed).
+  # Receives a datagram and returns the data and sender address information as
+  # [ data, [address_family, port, host, host] ], matching stdlib
+  # UDPSocket#recvfrom. Like the stdlib method, this blocks until a datagram is
+  # available and has no timeout of its own (see #timed_recvfrom for a variant
+  # that does). The host appears in both the hostname and numeric address
+  # positions; no reverse-DNS lookup is performed.
   #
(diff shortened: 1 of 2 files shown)


lib/rex/socket/udp.rb
```diff
@@ -117,37 +117,60 @@ def sendto(gram, peerhost, peerport, flags = 0)
...
+  #
+  def sender_addr_info(saddr)
+    af, host, port = Rex::Socket.from_sockaddr(saddr)
+    af_name = ::Socket.constants.grep(/^AF_/).find { |c| ::Socket.const_get(c) == af }.to_s
+    [ af_name, port, host, host ]
   end
+  private :sender_addr_info
 
   #
   # Calls recvfrom and only returns the data
```
sjanusz-r7: Not a blocker: This no longer calls to recvfrom
Seems to end up calling `sysread` under the hood unless I am mistaken.

spec/rex/socket/udp_spec.rb
```diff
@@ -0,0 +1,61 @@
+# -*- coding:binary -*-
+require 'spec_helper'
+
+RSpec.describe Rex::Socket::Udp do
+  let(:loopback) { '127.0.0.1' }
+
+  def make_server
+    Rex::Socket::Udp.create('LocalHost' => loopback, 'LocalPort' => 0)
```
sjanusz-r7: Not a blocker:
Thoughts on:
```suggestion
    described_class.create('LocalHost' => loopback, 'LocalPort' => 0)
```

sjanusz-r7 (approved):

sjanusz-r7 (approved):

sjanusz-r7 (approved):

PR #86, opened 2026-06-29
Catch ECONNRESET too for Windows

### LINKED ISSUE #85: Windows UDP behaviour issue
Seems like there's extra changes that need to be made, following up on the previous recv changes. On Windows, we hit:
```
Received Error: An existing connection was forcibly closed by the remote host. - recvfrom(2)
Received Error: An existing connection was forcibly closed by the remote host. - recvfrom(2)
```
On Linux, this isn't a problem.

zeroSteiner:
Catch ECONNRESET which is what Windows raises. This makes sure that `#timed_recvfrom` behaves on Windows as it does on Linux when the peer is unreacable and it's a "connected" UDP socket.

* Fixes #85
* Related to rapid7/metasploit-framework#21618

### DIFF
--- a/lib/rex/socket/udp.rb
+++ b/lib/rex/socket/udp.rb
@@ -169,7 +169,7 @@ module Rex::Socket::Udp
 
     data, saddr = recvfrom_nonblock(maxlen)
     [ data, sender_addr_info(saddr) ]
-  rescue ::Timeout::Error, ::Errno::ECONNREFUSED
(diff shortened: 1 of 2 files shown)


spec/rex/socket/udp_spec.rb.orig
adfoster-r7, line 1: Probably needs nuked 👀

sjanusz-r7:
Can merge once the `.orig` file is gone 👍

PR #87, opened 2026-06-29
Propagate the OS assigned port value (fixes LocalPort 0)

zeroSteiner:
Propagate the OS-assigned port value. This means when LocalPort is 0, you get the value that the OS assigned.

See this comment for context https://github.com/rapid7/metasploit-framework/issues/21604#issuecomment-4812169954

### DIFF
--- a/lib/rex/socket/comm/local.rb
+++ b/lib/rex/socket/comm/local.rb
@@ -273,10 +273,11 @@ class Rex::Socket::Comm::Local
         end
 
         sock.bind(Rex::Socket.to_sockaddr(ip, param.localport))
-
       rescue ::Errno::EADDRNOTAVAIL,::Errno::EADDRINUSE
         sock.close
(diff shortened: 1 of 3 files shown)


kx7m2qd:
core change in local.rb looks right, gating the localport update in the else branch means it only runs after a successful bind, and the specs check out for both tcp server and udp.

one thing I couldn't figure out though proxies.rb only has `require 'uri'` added, nothing else changed there. not obvious how that ties into propagating the os-assigned port, is it needed for something else in this PR or just something you noticed while working on this?

also since proxies.rb is touched, should there be a spec for LocalPort 0 going through a proxy too, or are the two direct-bind specs enough for what this PR is trying to fix?

asking because rapid7/metasploit-framework#21604 depends on this landing, wanted to actually understand what's changing before that work gets built on top of it

